PR #182, opened 2014-01-20
Add Kryo.setClassResolver(ClassResolver)

magro:
Add the possibility to set a `ClassResolver` after Kryo was constructed, but it's expected
that `Kryo.setClassResolver` is invoked before any (de)serialization is performed (other behaviour
is just not tested).

The Kryo instance is set on the new class resolver, so that the client does not have
to invoke this (also setters for ReferenceResolver and StreamFactory are adjusted to set Kryo).


magro:
I've implemented a builder based approach in msm, therefore obsolete.

PR #187, opened 2014-01-26
CompatibleFieldSerializer but duplicate field name will be accepted

bohrqiu:
Features like
com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer,but The
situation where a super class has a field with the same name as a
subclass will be accepted.


xfrendx:
I've the same problem now (using Kryo 3.0.3). "null" from super class attributes overrides same named attribute values. What's the status of this "bug" please?

magro:
@xfrendx The status is what you can read here. AFAICS the proposed solution replaces the existing issue with another one (if the superclass works on the field directly instead of using the getter), so it's not a real solution. Perhaps we should introduce a setting that allows to activate the new behavior, so that users have to actively choose this.
Apart from this the code needs some cleanup as it was mentioned.
If you want to push this forward go ahead, this would be great!

magro:
Resolved with #424.

PR #222, opened 2014-06-10
Allow easy multiple dimension class registration

Jyro117:
Added ability to easily register a type and all array types for that class up to a specified dimension. Example:

kryo.registerArrays(String.class, 3);

Will register String.class, String[].class, String[][].class, String[][][].class


NathanSweet:
I think it adds a bit too much clutter to Kryo's API. It would fit better in Kryo's Util class, or maybe the kryo-serializers project.

Issue #151, opened 2013-11-14
Move serializers for std jdk classes from kryo-serializers to kryo


magro:
Motivated by e.g. #88 and #100

megglos:
would love to see that :)

magro:
Do you want to submit a pull request? :-) Perhaps only for classes you're interested in...?

megglos:
If you like to, yes :)

Currently for me it's just UnmodifiableCollectionsSerializer, need that for getting a correct copy of object trees that contain Exceptions (Throwable#suppressedExceptions is by default an unmodifable list of size 0).

NathanSweet:
Given that this issue is 5 years old, it appears no one is interested in stepping up and bringing the serializers over. We can reopen should that happen, otherwise it is not difficult for people to get the serializers from there. Note that the serializers may need cleaning up to match the rest of the library, it should not just be a copy/pasta job. Besides, kryo-serializers needs an update to Kryo 5.0.0 before this could be done.

guidomedina:
So I'm guessing the wisest thing to do to minimize the work doing this would be wait for 5.0.0 is fully released, I can help with some modules but I can't promised to migrate the whole thing.
I would like though to start small, once 5.0.0 I can do the `joda-time` module which is the one I'm interested.
I wouldn't bother updating the project with all the serializers but rather module by module as the need arises.

NathanSweet:
Sounds reasonable!

isaki-x:
I've done a bit of work for some common Java types for my own use of Kryo, such as UUID, which might be worth considering as provided by default. I've also written a base serializer class for Guava multi-map serialization that is optimized for serialized data size, which I can also contribute should a Guava module be necessary (as well as handling for Immutable lists, sets, and multi-maps).

magro:
@guidomedina re waiting until 5.0.0: today I released 5.0.0-RC1 - I guess this is fine for you as well?

guidomedina:
yeah, I can start working with that, thanks ;-)

PR #284, opened 2015-02-04
Added UnmodifiableCollectionsSerializer to default config and added simp...

megglos:
https://github.com/EsotericSoftware/kryo/issues/151#


magro:
@NathanSweet WDYT?

NathanSweet:
I don't think JVM specific serializers should be registered by default. I guess including them is fine as long as it is documented.

magro:
Sounds reasonable. For jvm specific serializers the jdk compatibility should also be documented. kryo-serializers are built with jdk6-8 (openjdk/oracle) and ibm-jdk-7 (see https://jenkins.inoio.de/view/kryo/).

thomasdarimont:
Would be great to have that - whats the state of this PR?

groundhog2k:
@megglos Great PR!

PR #295, opened 2015-03-27
add type to newInstantiator in kryo

hepin1989:
may need more work ,for that we put the dependency in lib,which is an unmanaged lib


hepin1989:
@magro I will look into this deeper.

hepin1989:
have been fixed,will close,thanks.

hepin1989:
will update later,close this.

PR #297, opened 2015-04-01
Warn unregistered classes option

timotta:
With this option, unregistered classes are warned on kryos log instead of raise exception like the option registrationRequired do. 

This option is required mostly for processing large amount of data with third vendor libs, to see all classes that are not registered in the first running, instead of have to discover one class for each running. This happens with me using Spark and mlib.ALS

How to use:
```java
kryo.setWarnUnregisteredClasses(true)
```

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -123,6 +123,7 @@ public class Kryo {
 	private ClassLoader classLoader = getClass().getClassLoader();
 	private InstantiatorStrategy strategy = new DefaultInstantiatorStrategy();
 	private boolean registrationRequired;
+	private boolean warnUnregisteredClasses;
 
 	private int depth, maxDepth = Integer.MAX_VALUE;
 	private boolean autoReset = true;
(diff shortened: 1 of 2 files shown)


NathanSweet:
Do we need a setting for this? You could log a message from `ClassResolver#registerImplicit`.

brianfromoregon:
I have a similar feature locally, inside an overridden com.esotericsoftware.kryo.util.DefaultClassResolver#register


            if (warnImplicitRegistration && registration.getId() == NAME) {
                // Implicit registrations make for larger serialized forms.
                log.warn(String.format("Implicit registration: %s (%s)",
                        Util.className(registration.getType()),
                        registration.getSerializer().getClass().getName()));
            }

timotta:
@NathanSweet seems like a very needed feature. Are you suggesting to put this log even without the parameter? For me is even better, i just put the paramete to no change behaviour for previously users.

NathanSweet:
You can use ClassResolver#registerImplicit to log a message. Why do we need a setting on Kryo?

timotta:
just put the parameter to avoid changing previously behaviour, but if it is not necessary we can put the warn directly. Ok?

emlyn:
:+1: for this option. We use Kryo in Spark, it would be nice to be warned of unregistered classes without failing the whole run and only finding one at a time.
I'd rather not have to work out how to implement a different ClassResolver and make Spark use it if possible, an option like this would make it easier to use.

vidma:
good stuff. 

much useful for both:
- especially useful **in `Spark`, where `Kryo` is initialized internally**. so I doubt it would be easy to override the `DefaultClassResolver#register` without lots of copy-paste monkey-patching ;)
- more complex (i.e. normal) `scala` apps, where it's almost impossible to track down missing registrations one-by-one with hundred versions of collection classes like `::`, `ListOf1Element` and whatever else.

a config option would help too, but not necessarily, as e.g. `KryoRegistrator` in spark gives a reference to kryo instance where we can simply call: `kryo.setWarnUnregisteredClasses(true)` when needed.

so, dudes, it's sitting here for way  too long, just merge it ASAP :+1:  :)

vidma:
@NathanSweet

enragedginger:
+1 This would be nice to have.

Issue #224, opened 2014-07-01
Bug in Kryo#register(Class, int)


brianfromoregon:
Today register(Class, int) looks like this

	public Registration register (Class type, int id) {
		Registration registration = classResolver.getRegistration(type);
		if (registration != null) return registration;
		return register(type, getDefaultSerializer(type), id);
	}

I think it should be this

	public Registration register (Class type, int id) {
		return register(type, getDefaultSerializer(type), id);
	}

I have a data store where a class C is associated with multiple ids. I need to create a Kryo which can read this store, but if I can't register the same class with multiple ids then I have a problem. register(Class, Serializer, int) is already correct, register(Class, int) just needs to be changed to match.

magro:
Closing, reopen if still relevant.

brianfromoregon:
I view this as a bug still fwiw.

brianfromoregon:
Hi, I still think this is a valid bug, worth keeping open in case someone finds the cycles to fix?

Issue #288, opened 2015-02-22
Registering classes with same ID does not throw exception


szymex:
According to javadoc comment for Kryo.register(..) method, registering classes with same id should thrown exception, it does not happen. Below is a simple test:

```java
    public void testRegistrationWithSameId(){
        Kryo kryo = new Kryo();
        kryo.register(LinkedHashMap.class, 20);
        try {
            kryo.register(ArrayList.class, 20);
            fail("expected KryoException");
        } catch (KryoException ex){
            //expected
        }
    }
```

rcoh:
+1 for this issue.

NathanSweet:
The exception is only throw when `DEBUG` logging is on. @romix any idea why?

romix:
It doesn't look like we thrown any exception at all currently. The only thing I see is this:
```
https://github.com/EsotericSoftware/kryo/blob/master/src/com/esotericsoftware/kryo/Kryo.java#L445-L448
```

But we only issue a debug message here. No exceptions are thrown.

rcoh:
I ended up just adding a wrapper around Kryo to check if the id had been registered first. The code should probably throw an exception, but I suspect people are currently relying on the broken behavior.

magro:
@romix @NathanSweet Shall we commit on the new behaviour, or change it back to the previous one?

FYI, this change was introduced with commit 7c8bc3b329da6d2e0b5f2e325ad59325e70547c8:

> Changed registration of a different class with the same ID to a debug message.
> 
> The use case is in KryoNet's RMI, where we want to send the implementation of an object on the server as if it were the interface. This means the interface and the implementation use the same class ID.

NathanSweet:
Ah right. We should not throw an exception and the javadocs should not say one is thrown.

PR #317, opened 2015-06-09
Alters the javadocs for the various register(...) methods to more correctly match behavior

nonvirtualthunk:
A few of the register(...) methods had javadocs that did not exactly align with the behavior of their methods, I believe as a result of changes that happened while moving to kryo version 3. This pull request brings the javadocs back into line. If it would be more desirable to alter the code to match the docs I would be happy to do so, but I thought it more likely that the code held the source of truth.

The primary differences were:
1) calling register for a class that already has a registration will not override the previous registration, unless register(...) was called with an explicit serializer
2) calling register with an explicit ID that is already in use for a different type will no longer throw an exception, instead it logs at the debug level

I believe that this partially resolves issue #224 or at least makes the behavior in question predictable.

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -378,8 +378,8 @@ public class Kryo {
 	// --- Registration ---
 
 	/** Registers the class using the lowest, next available integer ID and the {@link Kryo#getDefaultSerializer(Class) default
-	 * serializer}. If the class is already registered, the existing entry is updated with the new serializer. Registering a
-	 * primitive also affects the corresponding primitive wrapper.
+	 * serializer}. If the class is already registered, no change will be made and the existing registration will be
+	 * returned. Registering a primitive also affects the corresponding primitive wrapper.
 	 * <p>
 	 * Because the ID assigned is affected by the IDs registered before it, the order classes are registered is important when
 	 * using this method. The order must be the same at deserialization as it was for serialization. */
@@ -389,9 +389,10 @@ public class Kryo {
 		return register(type, getDefaultSerializer(type));
 	}
 
-	/** Registers the class using the specified ID and the {@link Kryo#getDefaultSerializer(Class) default serializer}. If the ID is
-	 * already in use by the same type, the old entry is overwritten. If the ID is already in use by a different type, a
-	 * {@link KryoException} is thrown. Registering a primitive also affects the corresponding primitive wrapper.
+	/** Registers the class using the specified ID and the {@link Kryo#getDefaultSerializer(Class) default serializer}.
+	 * If the class is already registered this has no effect and the existing registration is returned. Providing an
+	 * ID that is already in use by a different type is unsupported and may result in undefined behavior. Registering a
+	 * primitive also affects the corresponding primitive wrapper.
 	 * <p>
 	 * IDs must be the same at deserialization as they were for serialization.
 	 * @param id Must be >= 0. Smaller IDs are serialized more efficiently. IDs 0-8 are used by default for primitive types and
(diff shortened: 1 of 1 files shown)


src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -389,9 +389,10 @@ public Registration register (Class type) {
 		return register(type, getDefaultSerializer(type));
 	}
 
-	/** Registers the class using the specified ID and the {@link Kryo#getDefaultSerializer(Class) default serializer}. If the ID is
-	 * already in use by the same type, the old entry is overwritten. If the ID is already in use by a different type, a
-	 * {@link KryoException} is thrown. Registering a primitive also affects the corresponding primitive wrapper.
+	/** Registers the class using the specified ID and the {@link Kryo#getDefaultSerializer(Class) default serializer}.
+	 * If the class is already registered this has no effect and the existing registration is returned. Providing an
+	 * ID that is already in use by a different type is unsupported and may result in undefined behavior. Registering a
```
magro: I'd say "Providing an ID that is already in use by a different type is unsupported and may result in undefined behavior." can be removed because this is already covered by the previous sentence "If the class is already registered this has no effect and the existing registration is returned.".

src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -417,9 +418,9 @@ public Registration register (Class type, Serializer serializer) {
 		return classResolver.register(new Registration(type, serializer, getNextRegistrationId()));
 	}
 
-	/** Registers the class using the specified ID and serializer. If the ID is already in use by the same type, the old entry is
-	 * overwritten. If the ID is already in use by a different type, a {@link KryoException} is thrown. Registering a primitive
-	 * also affects the corresponding primitive wrapper.
+	/** Registers the class using the specified ID and serializer. Providing an ID that is already in use by the same type
+	 * will cause the old entry to be overwritten. Providing an ID that is already in use by a different type is unsupported
+	 * and may cause undefined behavior. Registering a primitive also affects the corresponding primitive wrapper.
```
magro: Same here, I'd say that "Providing an ID that is already in use by a different type is unsupported and may cause undefined behavior." can be removed because it's already covered by the previous sentence "Providing an ID that is already in use by the same type will cause the old entry to be overwritten." (which is also correctly describing the implementation AFAICS).

src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -429,9 +430,9 @@ public Registration register (Class type, Serializer serializer, int id) {
 		return register(new Registration(type, serializer, id));
 	}
 
-	/** Stores the specified registration. If the ID is already in use by the same type, the old entry is overwritten. If the ID is
-	 * already in use by a different type, a {@link KryoException} is thrown. Registering a primitive also affects the
-	 * corresponding primitive wrapper.
+	/** Stores the specified registration. If the ID is already in use by the same type, the old entry is overwritten.
+	 * If the ID is already in use by a different type, undefined behavior may result. Registering a primitive also affects
```
magro: Again, I'd say "If the ID is already in use by a different type, undefined behavior may result." can be removed.

src/com/esotericsoftware/kryo/Kryo.java
nonvirtualthunk: I added that line because if you provide a type that has not been registered with an ID that has already been used it won't hit that initial check and will end up implicitly erasing whatever was previously registered with that ID. That is something of an edge case though, so it probably doesn't need to be in the javadocs. I'll remove it here (and in the following ones).

magro:
Also related to issue #288

magro:
Ok, let's wait with merging for the decision regarding the thrown KryoException (or just logging) as discussed in #288.

NathanSweet:
Thanks!

Issue #321, opened 2015-06-22
Is there any reason not to use names of enum?


kwon37xi:
When I see EnumSerializer, it does not use enum's name. It uses ordinal.

This makes a problem when the order of enum constants changed
When the order changed, if it throws an exception, it's ok.
But if the deserialized ordinal exists in the new enum class, kryo does not throw any exception, just sets wrong enum value.

I made EnumStringSerializer by myself, and want to contribute to this project.

But Kryo's enum support has been very long time. I'm not sure whether you have specific reason not to use names for a long time.

Is there any reason not to use enum's name?

NathanSweet:
Ordinals are smaller. We'd merge an Enum serializer that uses names, but it
won't be used by default. It should be called EnumNameSerializer.

magro:
@NathanSweet Don't we need a replaceDefaultSerializer then?

NathanSweet:
Don't think so, EnumSerializer is lower priority than any default
serializer added by a user.

magro:
Ah, right.

PR #323, opened 2015-06-23
Fix for #321 EnumNameSerializer

kwon37xi:
With EnumNameSerializer, enum values are serialized with their name.
This serializer prevents reading wrong values when enum's order changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 *~
 .*.swp
 .DS_STORE
+*.iml
 
 bin/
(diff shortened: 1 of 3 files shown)


romix:
Question: The names are written using writeString(). Wouldn't it be better to write them using writeObject? In this case, Kryo could re-use these names, if references are enabled. This could be useful for big graphs.

NathanSweet:
Good point, done.

PR #327, opened 2015-07-17
Fixed compile errors by reverting cast externalizable in Serializer

dorkbox:
Kryo in it's current state will not compile, since kryo.newInstance(type) returns type Object, and the implicit casting via generics does not take place. I only reverted the removal of casting made by change a0cb680 so that this class can now compile.
  
The specific error was:
ERROR
incompatible types
  required: java.io.Externalizable
  found:    java.lang.Object

Location: Line: 88:73
com/esotericsoftware/kryo/serializers/ExternalizableSerializer.java

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/ExternalizableSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/ExternalizableSerializer.java
@@ -85,7 +85,7 @@ public class ExternalizableSerializer extends Serializer {
 
 	 private Object readExternal (Kryo kryo, Input input, Class type) {
 		  try {
-				Externalizable object = kryo.newInstance(type);
+				Externalizable object = (Externalizable)kryo.newInstance(type);
 				object.readExternal(getObjectInput(kryo, input));
 				return object;
 		  } catch (ClassCastException e) {


NathanSweet:
Kryo compiles fine. `newInstance` uses generics so a cast is not needed.

dorkbox:
Even with generics, it strangely fails unless the explicit cast is present. I'll dig some more into the issue, likely something on my end.

magro:
@NathanSweet master is failing to compile with java 8, I'm getting

```

NathanSweet:
Sorry, OP was right. Must have been a compiler bug!

NathanSweet:
closes #327

dorkbox:
@magro when is the next planned release? very soon?

magro:
@dorkbox I should be able to cut a release tomorrow.

dorkbox:
@magro excellent. I await the release! (and thanks for confirming the compile bug)

magro:
@dorkbox Unfortunately the release is now blocked by issue #331

magro:
@dorkbox It's now released with kryo 3.0.3.

Issue #174, opened 2014-01-14
Support serialization of Kryo and Kryo serializers instances


romix:
It could be very useful if one could serialize instances of Kryo and serializers.  This would essentially provide a meta-information about Kryo's configuration. And this can be used to create/configure a Kryo instance on the deserializing side exactly in the same way without   any need for manual coding. 

Such a feature could be also interesting for use-cases, where Kryo is used as a representation for a long-term storage of data. In such scenarios, one typically needs to store some sort of meta-information (a "schema", so to say), so that later it is possible to determine, which format is used for the stored data. While this proposal does not completely solve this issue, it is probably a step in the right direction.

NathanSweet:
I understand the need, but I think with the provided serializers it's hard to serialize an entire, configured Kryo instance. I think I prefer not going down that road, at least for now.

PR #339, opened 2015-08-13
KryoBuilder, builder for Kryo's that implements KryoFactory, with its own Serializer

esialb:
One of the things I frequently find myself wishing for is a way to persist Kryo instances themselves, so that a blob of data can include not only the data but a Kryo used to read that data.  Modifying Kryo.java to be itself serializable in a useful way seems fruitless, but I struck across another idea that works: a serializable "builder" for kryos.  Instead of serializing a fully configured Kryo, we can serialize the steps used to configure a Kryo for use.

This pull request is for KryoBuilder (and the very small KryoBuilderTest) classes.  KryoBuilder implements KryoFactory, and provides KryoBuilder.KryoBuilderSerializer for its own serialization.

The source is not currently commented in any useful way, but I will be more than happy to document and clean it however necessary if this pull request is worth considering.

### DIFF
--- a/src/com/esotericsoftware/kryo/KryoBuilder.java
+++ b/src/com/esotericsoftware/kryo/KryoBuilder.java
@@ -12,29 +12,139 @@ import com.esotericsoftware.kryo.util.DefaultClassResolver;
 import com.esotericsoftware.kryo.util.DefaultStreamFactory;
 import com.esotericsoftware.kryo.util.MapReferenceResolver;
 
+/**
+ * Configurable {@link Kryo} builder (implementing {@link KryoFactory}) with a provided {@link Serializer}
+ * ({@link KryoBuilderSerializer}) to make the builder itself able to be serialized.<p>
+ * 
+ * Typical use case is for "self-descriptive" Kryo data.  One would create a {@link Kryo}
+ * instance using a {@link KryoBuilder}, then as the first object written, serialize the builder
+ * itself.  When deserializing, first deserialize the {@link KryoBuilder}, then from
+ * it build a {@link Kryo} to deserialize the remainder of the input.<p>
+ * 
+ * More advanced uses of {@link Kryo} may require serializing individual {@link Serializer} instances,
+ * such as if {@link #register(Class, Serializer)} or {@link #register(Class, Serializer, int)} is
+ * called on a {@link KryoBuilder} which must be serialized.  An alternative to serializing
+ * {@link Serializer}s is to subclass the {@link Serializer}, performing customizations in the
+ * constructor, and instead call {@link #addDefaultSerializer(Class, Class)}.<p>
+ * 
+ * {@link KryoBuilderSerializer} can serialize a {@link KryoBuilder}, but does not provide
+ * explicit support for potentially required objects, such as {@link Serializer} instances
+ * that must be registered.<p>
+ * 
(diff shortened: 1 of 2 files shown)


src/com/esotericsoftware/kryo/KryoBuilder.java
```diff
@@ -0,0 +1,362 @@
...
+			return object;
+		}
+		
+	}
+	
+	private KryoBuilder head;
+	private KryoBuilder tail;
+	private Step step;
+	private Object[] args;
+
```
magro: head, step and args can be final AFAICS, right?

magro:
@NathanSweet @romix LGTM, WDYT? 

Remember, before merging commits should be squashed.

NathanSweet:
I'm not sure. I've never needed to do this. In every nontrivial app I've always needed more customization than the builder provides.

esialb:
I tried to add any customization of the Kryo class itself; customizations of individual serializers might have to be accomplised by subclassing the serializer and configuring in the constructor.  i'm not sure of a way around this; adding serializer-specific builder customizations would require either significant additions to the builder on a per-serializer-class basis, or the use of some sort of interpreted language, storing customizations to be evaluated at runtime.

src/com/esotericsoftware/kryo/KryoBuilder.java
esialb: i think they can be made final, as long as KryoBuilderSerializer never needs to make backreferences; kryo.reference(object) currently happens before deserializing the tail step, and changing this would require removing the call to kryo.reference

esialb:
Individual Serializer instances can be stored with the builder, see

```java
public KryoBuilder register (Class type, Serializer serializer, int id)
```

but serializer serializers might have to be written

esialb:
Added a lot of comments to KryoBuilder.java

esialb:
rebased onto esotericsoftware/kryo.git master

NathanSweet:
> serializer serializers might have to be written

I have a feeling the builder for serializing Kryo state will only work for simple cases. I expect most apps will need to customize the Kryo instance or the registered serializers in ways that are difficult to serialize along with the Kryo state. I'm just not sure it's worth doing. I'm not sure there is a real world use case for this. I understand you want to get your objects back no matter how they were serialized, but I think maybe Kryo's very nature is to leverage class definitions and the Kryo configuration, making this difficult or impossible.

esialb:
Maybe this is a simplistic answer to the more general question of whether it is possible to use "schemas" with Kryo.  On further thought the goal is not to persist the state of the Kryo instance itself, or the serializers, but to persist a definition of how Kryo is serializing data.  The runtime-specific nature of Kryo configuration that you bring up suggests that trying to persist the Kryo (or its serializers) may not be a useful way to persist a schema for the serialized data, but it would be nice if it were possible somehow.

esialb:
Well, disregarding the question of persisting a KryoBuilder, this at least provides a handy implementation of KryoFactory.  Might be worth considering purely on that basis.

NathanSweet:
Related: #174. 

Kryo is only a framework for serializers. Serializers could be written to use schemas, though the ones that come with Kryo mostly rely on the Java class files for automatic serialization.

I understand the need, but I think with the provided serializers it's hard to serialize an entire, configured Kryo instance. I think I prefer not going down that road, at least for now.

Issue #341, opened 2015-08-23
Stackoverflow error due to parentScope of Generics being pushed as the same object


hierynomus:
In our product we're currently using Kryo 2.24.0 to serialize a data structure to another node. However in some specific cases during serialization a `StackOverflowError` occurs. with the following stacktrace:

```
java.lang.StackOverflowError
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:61)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
	at com.esotericsoftware.kryo.Generics.getConcreteClass(Generics.java:62)
[...]
```

After much debugging we see that in this case `kryo.pushGenericsScope(...)` is called with exactly the same `Generics` instance as that is already set. When this happens, the `parentScope` points to the same instance as the current scope, so you get a loop while resolving.

I've been able to create an isolated testcase in the following Gist: https://gist.github.com/hierynomus/a2f143a555466f83d9bc

This issue is reproducible with Kryo 3.0.3 when using the Gradle buildfile (Gradle 2.6) that is attached to the Gist.

In order to fix this, my initial thought was returning a boolean from `pushGenericsScope` whether or not it was pushed, and checking before popping that there was actually a push. And then inside the pushGenericsScope checking whether the same instance was being pushed. However this fails the _clirr_ checks. 

You could also do the same check before calling the `pushGenericsScope` / `popGenericsScope` combo. This will adhere to the _clirr_ checks.

I'd be happy to send a PR, but thought I'd discuss an approach first.

hierynomus:
Small update (and further pinpointing)...

Running this on MacOSX with _Java7 (1.7.0_71)_ gives a StackOverflowError
However running on _Java8 (1.8.0_25)_ runs successfully.

So the real question is what is different when switching the Java version?

hierynomus:
[...]

PR #342, opened 2015-08-24
Prevent creation of cyclic Generics stack when serializing specific generics object trees

hierynomus:
A proposed fix for #341.

To prevent the same `Generics` object becoming a parent in the tree of itself, no longer cache it, but rather calculate it on demand from the parameters already stored in the `FieldSerializer`.

This might have a small performance impact, but as of yet I haven't figured out a better way to fix it. Let me know whether you think this is the right fix.


magro:
@romix WDYT? I'm not into FieldSerializerGenericsTest and generics, but to me it looks good so far :-)

Regarding the StackOverflowTest I'd prefer to name it e.g. FieldSerializerGenericsTest with a test method that says what it's about. Perhaps there are already some other tests targeting FS and generics that could also go into that class.

romix:
I haven't had time to review it yet. Hopefully, I'll manage to find some time this week.
I really need to review this, because working with generics is a very subtle job ;-) 

I'm going to review the patch for both its correctness and performance impact. 

And at the first glance, it looks like it would introduce some additional overhead. The questions is: how much? We want to avoid too much overhead on generic-heavy object graphs...

magro:
@romix Great! :-)

hierynomus:
@magro I renamed the testcase as you suggested.

hierynomus:
I've got an alternative implementation in the following commit: https://github.com/hierynomus/kryo/commit/4b604d89158be09dd6d0dd73dd51efb0b7585c96

This does not incur the overhead of needing to recalculate the Generics object every time, but the downside is that it breaks binary compatibility

```
[ERROR] 7002: com.esotericsoftware.kryo.Generics: Method 'public Generics(com.esotericsoftware.kryo.Generics)' has been removed
[ERROR] 7002: com.esotericsoftware.kryo.Generics: Method 'public com.esotericsoftware.kryo.Generics getParentScope()' has been removed
[ERROR] 7002: com.esotericsoftware.kryo.Generics: Method 'public void resetParentScope()' has been removed
[ERROR] 7002: com.esotericsoftware.kryo.Generics: Method 'public void setParentScope(com.esotericsoftware.kryo.Generics)' has been removed
[ERROR] 7006: com.esotericsoftware.kryo.Kryo: Return type of method 'public com.esotericsoftware.kryo.Generics getGenericsScope()' has been changed to com.esotericsoftware.kryo.GenericsResolver
```

These are the best two solutions I could come up with that solve the problem at hand. Judging from the code, the binary incompatibility might not be so bad, as the `Generics` object is only used by the `FieldSerializer`/``FieldSerializerGenericsUtil`. Also all the other custom serializers in for instance:

- chill_scala: "com.twitter:chill_${scalaVersion}:0.7.0",
- kryo_scala: "com.github.romix.akka:akka-kryo-serialization_${scalaVersion}:0.3.3",
- kryo_serializers: "de.javakaffee:kryo-serializers:0.35"

Do not seem to use the `Generics` class or the associated changed method in `Kryo`.

But of course this is up to you. It does mean you need to increase the version a bit higher than only the patch number.

magro:
> Judging from the code, the binary incompatibility might not be so bad, as
the Generics object is only used by the
FieldSerializer/`FieldSerializerGenericsUtil. Also all the other custom
serializers in for instance:
>
> chill_scala: "com.twitter:chill_${scalaVersion}:0.7.0",
> kryo_scala:
"com.github.romix.akka:akka-kryo-serialization_${scalaVersion}:0.3.3",
> kryo_serializers: "de.javakaffee:kryo-serializers:0.35"
>
> Do not seem to use the Generics class or the associated changed method in
Kryo.

IMHO we should go through the code and make classes/methods package private
that are not intended to be part of the public api. If we do that before,
we can choose this change here as it's then no longer affecting the public
api and we're safer in the future. I'll create an issue for that.

hierynomus:
I do agree, I don't think the Generics object should have been exposed as a public API. If you prefer the alternative commit over the first one, I'll merge that into this PR, and revert the first change which added the runtime overhead. Just let me know.

magro:
I think this makes sense as soon as #343 is solved. I'm currently/this week on vacation, therefore I'll not be able to work on 343 the next days.

hierynomus:
OK, merged in the other commit and reverted the changes to the production code of the first version. I could take a stab at #343 if you want.

magro:
Am 26.08.2015 15:39 schrieb "Jeroen van Erp" <notifications@github.com>:
>
> OK, merged in the other commit and reverted the changes to the production
code of the first version. I could take a stab at #343 if you want.

Yes, that would be great! :-)

src/com/esotericsoftware/kryo/GenericsResolver.java
```diff
@@ -0,0 +1,34 @@
+package com.esotericsoftware.kryo;
+
```
romix: Please add the standard license header to any new file you introduced.

src/com/esotericsoftware/kryo/GenericsResolver.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/** Helper class that resolves a type name variable to a concrete class using the current class serialization stack
+ *
+ * @author Jeroen van Erp <jeroen@hierynomus.com> */
+public class GenericsResolver {
```
romix: This class is currently declared public. Is GenericResolver going to be exposed to the users? 
If yes, why and how? What would be the API?
If not, it should not be public I guess.

src/com/esotericsoftware/kryo/GenericsResolver.java
```diff
@@ -0,0 +1,34 @@
...
+    private Deque<Generics> stack = new ArrayDeque<Generics>();
+
```
romix: Deque was introduced in Java7 only. Do we really need to use Deqeue here?
Or could a vector/list/etc be used instead? After all, we only need to mimic a stack, right?

src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -1161,22 +1161,15 @@ public boolean isClousre (Class type) {
...
 	public void popGenericsScope () {
-		Generics oldScope = genericsScope;
-		if (genericsScope != null) genericsScope = genericsScope.getParentScope();
-		if (oldScope != null) oldScope.resetParentScope();
+		genericsResolver.popScope();
 	}
 
-	public Generics getGenericsScope () {
-		return genericsScope;
+	public GenericsResolver getGenericsScope () {
```
romix: I'm a bit unhappy about the fact that `getGenericScope` returns a value of type `GenericsResolver`. I think it would be better if you'd rename the method into `getGenericResolver`, similar to what we do for other resolvers in Kryo.

romix:
Overall, I think the latest  version of your PR is OK. In the initial version, you re-computed  the scopes on the hot path inside `read`/`write` methods, which is a bad idea. But now you do it only during the serializer construction, which is the right thing to do.

src/com/esotericsoftware/kryo/GenericsResolver.java
hierynomus: Yes, I would agree, _but_ `FieldSerializer` and `FieldSerializerGenericsUtil` live in a different package than `Kryo`, so the class itself needs to be public. I'll restrict the methods in there.

src/com/esotericsoftware/kryo/GenericsResolver.java
hierynomus: Actually Java6. If we want to maintain Java5 compatibility, a `LinkedList` might be best.

romix:
Looks good. Thanks!

romix:
@magro Martin, could you check why our continuous build system fails because of clirr?

magro:
According to the comments above these are incompatible changes, therefore it's expected that clirr complains. This should also be the case locally.

One way to handle this is to list  the incompatible changes somewhere in the clirr plugin config in the pom.xml or disable clirr until the next release.
I'm still on vacation this week, I could take care of this afterwards. 

Btw, this PR contained multiple commits, it would've been better to squash them to a single commit because otherwise preparing release notes / changes gets unnecessarily hard - and the git history gets bloated.

romix:
> Btw, this PR contained multiple commits, it would've been better to squash them to a single commit 
> because otherwise preparing release notes / changes gets unnecessarily hard - and the git history gets 
> bloated.

You are right! Sorry. I'll do it like this next time. 
I wish Github would have some support for this in its Web GUI.

magro:
Yeah, that's right. In other projects often pull requesters are asked to squash commits after things are sorted out.

zEvg:
Any chance this fix going to be released any time soon ? (3.1.0 ?)
This issue drives us crazy. Is there any workaround except building against 3.1.0-SNAPSHOT ?

magro:
4.0 should be available the next days.

magro:
We just released 4.0.0.

PR #344, opened 2015-08-28
Propagate flush to underlying stream

snicolai:
Adjust incorrect test cases.


magro:
@NathanSweet I think we can merge this, or do you have any objections?

NathanSweet:
It takes away the ability to flush the Output without flushing the stream, but I doubt anyone relies on that.

Issue #360, opened 2015-10-29
Support java 8 specific serializers


magro:
Kryo should support some java 8 specific serializers, like e.g. for the new DateTime or Optional. To do this we need to adjust the build, and I think we should do the following:
- compile sources and tests with jdk8 (so that java 8 specific classes are available), still with target version 1.6
- run tests with several jdks: 1.6, 1.7, 1.8, ibm etc, while the previously compiled classes and test classes are used
- for jdks < 1.8 tests for java 8 specific serializers would be excluded 

This means that java 8 specific serializers would be located in the regular source path, kryo can only be compiled with jdk8. At runtime the default registration of some java 8 specific serializer would have to check the java version (unsurprisingly).

@NathanSweet @romix WDYT?

magro:
@NathanSweet @romix Do you have any objections to require jdk8 for building kryo?

magro:
Done

PR #362, opened 2015-11-01
Add java8 support and serializer for java.util.Optional

magro:
As said in #360 we should add support for java 8 specific serializers. I've added a "showcase" serializer for java 8's `Optional`, and added a build profile that excludes tests for java8 specific serializers for jdks < 1.8 (such tests must be located in the package `**/java8/`).

To verify that it's working as intended one can compile/run tests with jdk 8 with

    $ mvn -f pom-main.xml test

(which should report 157 successful tests)

and afterwards run tests with jdk < 1.8 with

    $ mvn -f pom-main.xml surefire:test

(which should report 154 successful tests).

Before/after the merge of this PR I need to change our CI build (https://jenkins.inoio.de/job/kryo/)  accordingly - build with jdk 8, only run tests with other jdks.


magro:
@NathanSweet @romix WDYT? Do you have any objections to require jdk8 for building kryo?

romix:
I don't have a strong opinion here. If our users are not affected by this, I've nothing against requiring jdk8.

PR #367, opened 2015-11-13
Non null serializer

mr-pathak:
A serializer which skips storing null field. Its similar to `CompatibleSerializer`, just that instead of storing a byte for null values it skips them. This is achieved by appending a bit set header.


mr-pathak:
@NathanSweet  This is to add a new serializer which doesn't store null bytes and is similar to CompatibbleSerializer.

magro:
Would it make sense to make this feature a setting on CompatibleFieldSerializer?

NathanSweet:
It doesn't make sense to avoid writing fields this way. Eg, while an uninitialized int may be 0, a class may initialize it to something else. To do it properly you'd need to create an instance solely to determine the default values for that specific class (I do this in [JsonBeans](https://github.com/EsotericSoftware/jsonbeans/)).

I agree it makes most sense as an option for CompatibleFieldSerializer.

PR #368, opened 2015-11-16
Try to load classes with fallback that uses current ClassLoader.

cerko2:
If running in an osgi environment the class loader that started kryo
can differ from the one that contains the class we want to load.
This introduces a fallback when class loading fails that
attempts to load the class with an osgi provided class loader.


magro:
Thanks for the PR!

Issue #377, opened 2015-12-17
Kryo does not correctly support parameterized type hierarchy


mathieucarbou:
__Test case:__

```
@RunWith(JUnit4.class)
public class SerializationTest {

  @Test
  public void testParameterizedTypesSupport() throws Exception {
    LongHolder o1 = new LongHolder(1L);
    LongListHolder o2 = new LongListHolder(asList(1L));

    Kryo kryo = new Kryo();
    Output buffer = new Output(512, 4048);
    kryo.writeClassAndObject(buffer, o1);
    kryo.writeClassAndObject(buffer, o2);
  }

  interface Holder<V> {
    V getValue();
  }

  static abstract class AbstractValueHolder<V> implements Holder<V> {
    private final V value;

    AbstractValueHolder(V value) {
      this.value = value;
    }

    @Override
    public V getValue() {
      return value;
    }
  }

  static abstract class AbstractValueListHolder<V> extends AbstractValueHolder<List<V>> {
    AbstractValueListHolder(List<V> value) {
      super(value);
    }
  }

  static class LongHolder extends AbstractValueHolder<Long> {
    LongHolder(Long value) {
      super(value);
    }
  }

  static class LongListHolder extends AbstractValueListHolder<Long> {
    LongListHolder(List<Long> value) {
      super(value);
    }
  }
}
```

__Error:__

```
com.esotericsoftware.kryo.KryoException: java.lang.ClassCastException: java.util.Arrays$ArrayList cannot be cast to java.lang.Long
Serialization trace:
value (org.terracotta.management.SerializationTest$LongListHolder)

	at com.esotericsoftware.kryo.serializers.ObjectField.write(ObjectField.java:101)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.write(FieldSerializer.java:518)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:628)
	at org.terracotta.management.SerializationTest.testParameterizedTypesSupport(SerializationTest.java:32)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
	at java.lang.reflect.Method.invoke(Method.java:597)
[...]

Issue #384, opened 2016-02-12
ClassCastException When Writing a Map After Previously Reading a Map with Different Generic Type


jlmorton:
While reading an object with a generic type map, Kryo will create a MapSerializer and set the generic type on the serializer.  When the FieldSerializer encounters a reference to the same map, it will again set the generic types of the MapSerializer, but will return without re-setting the generic types.

After this, a subsequent request to serialize a map with different generic types different will lead to a ClassCastException, as demonstrated in the test case below:

    @Test
    public void testSerializeMapAfterDeserialization() throws IOException {
        Kryo kryo = new Kryo();

        byte[] serializedMapHolder = serialize(kryo, new MapHolder());
        deserialize(kryo, serializedMapHolder);

        Map<Integer, List<String>> stringMap = new HashMap<>();
        stringMap.put(1, new ArrayList<>());
        serialize(kryo, stringMap);
    }

    static class MapHolder {
        private Map<Integer, String> mapOne = new HashMap<>();
        private Map<Integer, String> mapTwo = this.mapOne;
    }

    private byte[] serialize(Kryo kryo, Object object) throws IOException {
        try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream(); Output output = new Output(outputStream)) {
            kryo.writeClassAndObject(output, object);
            output.flush();
            return outputStream.toByteArray();
        }
    }

    private Object deserialize(Kryo kryo, byte[] serializedBytes) throws IOException {
        try (ByteArrayInputStream inputStream = new ByteArrayInputStream(serializedBytes); Input input = new Input(inputStream)) {
            return kryo.readClassAndObject(input);
        }
    }

This test case results in the following error:

java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.lang.String

	at com.esotericsoftware.kryo.serializers.DefaultSerializers$StringSerializer.write(DefaultSerializers.java:184)
	at com.esotericsoftware.kryo.Kryo.writeObjectOrNull(Kryo.java:603)
	at com.esotericsoftware.kryo.serializers.MapSerializer.write(MapSerializer.java:109)
	at com.esotericsoftware.kryo.serializers.MapSerializer.write(MapSerializer.java:39)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:625)
	at com.test.MapSerializerTest.serialize(MapSerializerTest.java:36)
[...]

PR #385, opened 2016-02-12
[BugFix] Reset the serializer generics when returning a previously read object

jlmorton:
This pull request fixes the bug described in issue: https://github.com/EsotericSoftware/kryo/issues/384

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -704,7 +704,10 @@ public class Kryo {
 			T object;
 			if (references) {
 				int stackSize = readReferenceOrNull(input, type, false);
-				if (stackSize == REF) return (T)readObject;
+				if (stackSize == REF) {
+					serializer.setGenerics(this, null);
+					return (T)readObject;
+				}
 				object = (T)serializer.read(this, input, type);
 				if (stackSize == readReferenceIds.size) reference(object);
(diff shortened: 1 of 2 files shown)


magro:
@jmorton Can you add the test from the issue to the commit?

@romix WDYT?

magro:
@jmorton Does this fix #377 as well?

jlmorton:
Thanks, I've added the test to this pull request.  I wasn't quite sure where it best fit, but I added it to MapSerializerTest.

I did notice #377 before I opened #384.  Interestingly, this change did in fact fix #377 in 3.0.1, which is the version we are currently using.  However, it did not fix it in master.  I've spent some time investigating #377 and I think I see an additional issue in master that is breaking #384.  I don't have a fix at the moment, but I'll post some thoughts on that thread.

magro:
@romix Can you have a look at this one and #377? Otherwise I'd have to understand FieldSerializer/Generics details which would take some time (and not sure when I could start looking at this).

romix:
@magro I'll try to have a look. But I'm very busy at work recently. And this kind of bugs is usually very subtle and requires quite some attention and time to be analysed and fixed properly. So, I promise to have a look, but I don't promise it will happen very soon ;-)

magro:
@romix Ok :-)

magro:
Resolved by #433

PR #388, opened 2016-02-21
Resolve #386: Automatically test serialization compatibility

### LINKED ISSUE #386: Automatically test serialization compatibility
... for common / provided serializers, and for both standard IO and Unsafe-based IO.

Here's a first attempt: [ser-compat-check](/magro/kryo/commits/ser-compat-check)

magro:
Tests serialization compatibility for the various Input/Output
variants: an instance of `TestData` was serialized to a file
(one for each IO variant), which is deserialized and compared
to a new `TestData` instance (the expected value).

The TestData class contains fields for each default serializer
(respectively the related type) and some arbitrary data structure
(`Person`) that shall mimick some real world data (perhaps not
really needed, just to have a class that's not too trivial).

If a new default serializer is added, the `TestData` should be
altered, and data files should/must be recreated (just delete
the existing test data and run the test to create new files).

Implementation notes:

* Needed to override KryoTestCase.assertEquals because the standard
  junit assert equals does not handle object graphs accordingly.
  commons EqualsBuilder.reflectionEquals is also not sufficient, therefore
  I added a reflectionEquals I used in kryo-serializers and memcached-
  session-manager and in other projects (so it's fairly good tested).
  Also tried unitils' `ReflectionAssert.assertReflectionEquals`, but
  this failed for StringBuilder/StringBuffer.


magro:
@NathanSweet @romix I'd be happy if you could have a look at this. If not it's ok as well :-)

romix:
@magro Overall, the direction looks good. But I think you need to provide more comments and descriptions.

For example:
- How the overall process of testing should look like: we have a set of "canonical" tests files to test against, we always check if we can deserialize their contents, etc.
- How do you produce the *.ser data files? What do you need to run to produce them?
- When these files are supposed to be produced? E.g. every new release, every new major release, etc...
- What about unsafe? Unsafe produces output which is target architecture dependent. We probably need to handle it somehow, i.e. we should have different *.ser files for different popular architectures.
- How the evolution process looks like: If we want to test some new aspects that were not covered by the original payload class, do we add new fields to this class? Do we add new payload classes into a specific directory, etc.

magro:
@romix Thanks for your valuable input, I'll improve comments.

Regarding the target architecture, we can only test what we have in our CI environment (debian x86_64), otherwise we'd have to invest more work into the CI env. This might be done in the future, for now I'd start with what we have.
I'm currently also wondering what exactly "target architecture" is referring to and what we should use to determine the "target architecture". I assume that we need to know the underlying cpu architecture and not the bitness of the jvm (which is what `System.getProperty("os.arch")` returns according to [this post](http://mark.koli.ch/javas-osarch-system-property-is-the-bitness-of-the-jre-not-the-operating-system)). Do you have a quick suggestion how we should determine the target architecture?

romix:
Regarding the target arch: I think we could check it dynamically. We could just serialize an integer using Unsafe and then check if it was serialized big or little endian approach. I think that the enidanness is the only representation difference between different JVM's methods of unsafe serialization. But I'm not 100% sure. May be JVM's have also some other characteristics that would result in different serialization?

magro:
Ok, we could probably also use
https://docs.oracle.com/javase/7/docs/api/java/nio/ByteOrder.html to check
endianness.

Not sure if there are other things influencing unsafe serialization.
Am 22.02.2016 16:06 schrieb "romix" <notifications@github.com>:

romix:
Yes, using the ByteOrder is a good idea!

romix:
Regarding the documentation: May be some of it should also be copied into Readme.txt, so that people know how and when to add/change *.ser files.

magro:
@romix I updated files to be endianness specific. Regarding docs: I'd say this information is very specific should be relevant for very few people so that I wouldn't want to clutter the main documentation with this. Instead I added a log line for the error case that points the user to the class javadoc. WDYT?

magro:
Btw, I've added separate commits, I'd squash them before merging, when everything's fine...

romix:
Regarding docs: I guess you are right. Log line is fine.

magro:
@romix The test is now mentioned in the README (in versioning/upgrading), and I squashed commits. Ok to merge?

romix:
@magro Yes, merge it!

magro:
Thanks, @romix for review!

PR #391, opened 2016-02-22
Resolve #389  ByteBufferOutput writeVar*() changes endianness underlying nioBuffer + minor fixes

### LINKED ISSUE #389: ByteBufferOutput writeVar*() changes endianness underlying nioBuffer
Calling writeLong()/writeInt() after writeVarLong()/writeVarInt() will write the Long/Int in little endian order, because the order() in the nioBuffer is set to little endian in writeVar*().
A certain use-case would be having a fixed-size header which contains the number of elements in the buffer, which will be written to after writing the VarLong/VarInt. 

Perhaps the order() of the nioBuffer should be reverted to the original order after writeVar*() ?

stau:
Tests: 
Each of the bottom cases 

Fixes: 
- ByteBufferInput
setBuffer did not take the byteOrder from ByteBuffer
setPosition did not set the position in ByteBuffer
skip did not skip bytes in ByteBuffer
setLimit did not set limit in ByteBuffer

- ByteBufferOutput
during writeVar*() a KryoException caused by require(int) did not reverse the byteOrder
order did not set order in ByteBuffer
setPosition did not set position in ByteBuffer

@romix 
Please take a look when you have time :)

### DIFF
--- a/src/com/esotericsoftware/kryo/io/ByteBufferInput.java
+++ b/src/com/esotericsoftware/kryo/io/ByteBufferInput.java
@@ -109,6 +109,7 @@ public class ByteBufferInput extends Input {
 		position = buffer.position();
 		limit = buffer.limit();
 		capacity = buffer.capacity();
+		byteOrder = buffer.order();
 		total = 0;
 		inputStream = null;
 	}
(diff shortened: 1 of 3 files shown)


romix:
I haven't looked in depth yet. But I noticed one thing: you try to fix two different issues in this PR:
- byte order issue
- position-handling issue.

I think we should not mix them. Could you please split this PR into two PRs, each solving only one specific issue? Thanks!

stau:
I have split this PR into two different PRs:
https://github.com/EsotericSoftware/kryo/pull/393
https://github.com/EsotericSoftware/kryo/pull/394

PR #393, opened 2016-02-23
Fix #389 ByteBufferOutput: revert ByteOrder on overflow in require during writeVar*()

### LINKED ISSUE #389: ByteBufferOutput writeVar*() changes endianness underlying nioBuffer

stau:
Test:
A case for when writing 
 - varInt causes an overflow in the buffer
 - varLong causes an overflow in the buffer

during writeVar*() a KryoException caused by require(int) did not revert the byteOrder

### DIFF
--- a/src/com/esotericsoftware/kryo/io/ByteBufferOutput.java
+++ b/src/com/esotericsoftware/kryo/io/ByteBufferOutput.java
@@ -200,12 +200,16 @@ public class ByteBufferOutput extends Output {
 	/** @return true if the buffer has been resized. */
 	protected boolean require (int required) throws KryoException {
 		if (capacity - position >= required) return false;
-		if (required > maxCapacity)
+		if (required > maxCapacity) {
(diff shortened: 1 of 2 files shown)


magro:
@stau After the other PR is merged this one now has conflicts with the test case. Can you please resolve it?

@romix Same here, LGTM, what do you say?

stau:
@magro The test case conflicts have been resolved

magro:
@stau :+1:

romix:
LGTM.

PR #394, opened 2016-02-23
Fix positioning 

stau:
Fix positioning and setting of byteOrder on setting buffer + tests

The tests for the ByteBufferInput are aimed for externally given buffers.


magro:
The code looks good to me. The commit message has as first line "- ByteBufferInput", which is not that useful in a list of commits, can you use s.th. more meaningful? Thanks!

stau:
Thanks for the review, I've changed the first line of the commit into something more meaningful.

magro:
Thanks @stau! @romix LGTM, do you want to have a look at it as well, as you know this IO stuff better? Otherwise I'd merge it.

romix:
LGTM.

PR #396, opened 2016-03-01
Adds Charset serializer

### LINKED ISSUE #364: Few of the java Charsets classes from rt.jar can not be deserialized
Charsets like sun.nio.cs.UTF_8,  UTF-16, UTF-16BE, UTF-16LE, ISO_8859_1 are defined package or private scope and are unable to be deserialized. These classes may be used internally by java objects and without user knowing it.

Is it a good idea to supply serializer for these Charsets? Serializer code is very simple but may be very useful and moreover streamlined for all kryo users. Please let me know if its a good idea or not. If yes then I will submit PR.

magro:
Is added as default serializer for java >= 7.

If registration is set to required the user has to register the concrete charset class, e.g. with `kryo.register(StandardCharsets.ISO_8859_1.getClass());`.

JavaVersion and the whole java8 package should be removed/moved up to serializers in a separate commit.

Resolves #364


magro:
Note2self: Test should go to DefaultSerializersTest, SerializationCompatTest must be updated.

magro:
The check for java 7 in the serializer should not be necessary.

PR #399, opened 2016-03-02
Update README.md to use the new CollectionSerializer.setElementClass(..)

yonghoon:
`CollectionSerializer.setElementClass(Class elementClass` is obsolete. Change it to use `CollectionSerializer.setElementClass(Class elementClass, Serializer serializer)`


magro:
Thanks for the PR!

Issue #382, opened 2016-02-02
Kryo breaks down while serializing genomic data


mulugetam:
**Bug**:
While running a genomic pipeline workload over a Spark cluster, we found out that Kryo breaks down throwing a `java.lang.NegativeArraySizeException` exception. The stack trace indicates that it happens when Spark attempts to blockify our genomic object using Kryo.

**Cause**:
The stash in the Cuckoo hash implementation of `IdentityObjectIntMap `fills up rapidly initiating a doubling of the hash table capacity beyond `Integer.MAX_VALUE`. 

We were able to isolate the problem object and serialize it out (using Java serialization) to a file of size 743M and reproduce the problem independently of our running context (We wrote a simple test application that de-serializes our object file using Java serialization and then attempt to serialize it using Kryo). 

The memory footprint of our genomic object (https://github.com/DimitrisAndreou/memory-measurer):

- size: 1499106912 bytes (1.4G)
- # objects: 57,652,339
- # primitives: int x 57688824, boolean x 57652108, char x 74
- # references: 86,478,445

Further debugging reveals that the stash size fills up for a load factor as low as 5%:

**Rudimentary work around**:
The rudimentary work around we took to temporarily solve the problem is to gradually increase the _stashCapacity_ based on the stash load factor.

_Current implementation_:

```
private void resize (int newSize) {
…
stashCapacity = Math.max(3, (int)Math.ceil(Math.log(capacity)) * 2); 
… 
}
```

_Changed to_:
 
```
private void resize (int newSize) {
…
stashCapacity = Math.max(3,(int)(Math.ceil(Math.log(capacity)) * (2.0 + stashSize/stashCapacity))); 
… 
}
```

In the worst and most extreme case the change would increase the `stashCapacity `from about 44 elements (current implementation) to about 65 elements. And this has temporarily fixed the bug for us.

Attached you may find the raw data from our runs using different load factors (all of which break down for the current implementation). Also, since Kryo is being used to blockify other objects inside Spark, it should NOT break for the default load factor (0.8) and initial capacity (32). 

[kryo_bug_raw.xlsx](https://github.com/EsotericSoftware/kryo/files/113379/kryo_bug_raw.xlsx)

Thanks,
Mulugeta

magro:
@mulugetam Great analysis! Do you want to submit a pull request?

magro:
@mulugetam Is this still an issue for you or shall we close this?

mmammo:
[...]

PR #400, opened 2016-03-02
Modified utility hash maps to gracefully handle very large number of …

mulugetam:
Modified the util/hash maps to gracefully handle very large number of entries (>= 2^16) and resolve the bug described #382 

The rationale behind the change as well as the jmh performance measurements are in #382

### DIFF
--- a/src/com/esotericsoftware/kryo/util/IdentityMap.java
+++ b/src/com/esotericsoftware/kryo/util/IdentityMap.java
@@ -20,18 +20,20 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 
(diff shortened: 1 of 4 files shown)


magro:
@mulugetam Awesome, thanks!

romix:
@mulugetam Is there any performance impact? Does it make maps with < 2^16 slower? If so, by how much?

mulugetam:
@romix no it doesn't make it slower (within margin of error). The performance impact for hash maps < 2^16 is the addition of an if condition; which I compensated for by avoiding a _getfield opcode_ which was partly not implemented in the original source.The details of my jmh performance measurements are in #382

romix:
@mulugetam Thanks for the JMH performance measurements! And thanks for this PR. Well done!

PR #402, opened 2016-03-09
make FieldSerializer.serializeTransient usable

leewin12:
FieldSerializer.serializeTransient is exist but not usable.
So just let user config serializeTransient to serialize transient fields.


magro:
Can you update against current master? Configuration changed with #414. Thanks!

PR #365, opened 2015-11-10
Add forward compatibility to TaggedFieldSerializer in issue #352.

levyfan:
Add forward compatibility to TaggedFieldSerializer in issue #352.

Add ignoreUnknownTags flag in Kryo configuration. If it is set, TaggedFieldSerializer will ignore unknown field tags in input stream. Otherwise a KryoException will be thrown.

PR #414, opened 2016-03-28
Introduce FieldSerializerConfig to encapsulate config in Kryo

magro:
(As adressed on the mailing list: https://groups.google.com/d/msg/kryo-users/G16daCijbWo/1igHI7WcBgAJ)

The `Kryo` class is more and more used for "global" configuration of serializers like `FieldSerializer` etc (here's another PR: #365). The semantics of such settings and especially for which serializer it's applied can often only be clarified via documentation.

This change introduces a `FieldSerializerConfig` class that encapsulates configuration for `FieldSerializer`. It allows to configure settings for all `FieldSerializer` instances created via `Kryo.getFieldSerializerConfig`. On each `FieldSerializer` the config can be changed/overridden, this is done directly on the serializer (instead e.g. requiring to invoke `fieldSerializer.getConfig.setSomeThing`) to stay binary compatible and because forcing to use the config here does not seem to be of much value.
The `FieldSerializer` internally uses the config object so that any further settings are directly added to `FieldSerializerConfig` and there's only a single place to keep config settings.

`Kryo.asmEnabled` might be used by different serializers, but because until now it was only used by `FieldSerializer` this setting is also kept in `FieldSerializerConfig`. `Kryo.set/getAsmEnabled` is marked as deprecated with a hint to use `getFieldSerializerConfig` instead - this is done to stay binary compatible.

`Kryo.setCopyTransient` is removed because it was introduced after the latest release 3.0.3 and therefore not yet a published/released. Now `Kryo.getFieldSerializerConfig.setCopyTransient` has to be used.

If this is considered to be merged then #365 should also follow this approach.

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -45,6 +45,7 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 
 import com.esotericsoftware.kryo.serializers.DefaultSerializers.URLSerializer;
+import com.esotericsoftware.kryo.serializers.FieldSerializerConfig;
 import com.esotericsoftware.kryo.serializers.OptionalSerializers;
 import com.esotericsoftware.kryo.serializers.GenericsResolver;
 import com.esotericsoftware.kryo.serializers.TimeSerializers;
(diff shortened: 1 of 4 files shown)


romix:
Overall, LGTM. 

Don't you want to make the fields of the new configuration class private? After all you provide getters/setter anyways.

Another thing: I totally understand why this new class is inside the`FieldSerializer` class. On the other hand, `FieldSerializer` is pretty big already. So, may be the new class should be in its own file, outside of `FieldSerializer`?

magro:
@romix Thanks for your review! I changed things according to your review, everything makes sense :-)

romix:
@magro LGTM

Issue #215, opened 2014-05-03
Java8 Lambdas cannot be deserialized


ikabiljo:
Deserializing Java8 lambdas fails with ClassNotFoundException:
Test$$Lambda$1/1279149968

Example:

```
  private <T> T kryoSerDeser(T r) throws IOException, ClassNotFoundException {
    Kryo kryo = new Kryo();
    ByteArrayOutputStream baos = new ByteArrayOutputStream();
    try (Output output = new Output(baos)) {
      kryo.writeClassAndObject(output, r);
    }

    try (Input input = new Input((new ByteArrayInputStream(baos.toByteArray())))) {
      return (T) kryo.readClassAndObject(input);
    }
  }

  @Test
  public void testLambdaInKryo() throws IOException, ClassNotFoundException {
    Runnable r = () -> System.out.println("works");
    kryoSerDeser(r).run();
  }

```

Using Java's standard ObjectOutput/Input streams on lambdas implementing Serializable interface works fine.

rgilles:
Hi Kryo,
I would like to know if there is any way to serialize lambda that are not casted with a & Serializable.
By the way. I did a test on the 3.0.1 and I was oblige to do the following in order to make it works.
```
    @Test
    public void serializeSerializableLambdaWithKryo() throws Exception {
        Callable<Boolean> doNothing = (Callable<Boolean> & Serializable)(() -> true);
        Kryo kryo = new Kryo();
        kryo.setInstantiatorStrategy(new Kryo.DefaultInstantiatorStrategy(
                new StdInstantiatorStrategy()));
        kryo.register(java.lang.invoke.SerializedLambda.class);
        kryo.register(Class.forName(Kryo.class.getName() + "$Closure"), new ClosureSerializer());
        Callable<Boolean> result = serialize(kryo, doNothing);
        assertThat(result.call(), is(true));
    }
```
As you can see the Closure class is private into the kryo class. therefore you need to load it manually.
Maybe an update of the documentation?

Regards,

Romain.

orionll:
+1. Please, provide a full example for serializing/deserializing a lambda on the main page.

magro:
Related: #299

PR #415, opened 2016-03-29
Make Closure public, moving it to ClosureSerializer

### LINKED ISSUE #299: Can't register ClosureSerializer
With change 1c5562d users are forced to register `ClosureSerializer` on their own.
The problem is that it is not possible since `Closure.class` is private.

magro:
Closure is moved to ClosureSerializer because they're tightly coupled and this makes the connection more obvious.

This also renames `Kryo.isClousre` to `isClosure`, assuming that it was not used until now it's ignored regarding binary compatibility.

A test for closure serialization is added together with the required maven foo for compiler settings.

Fixes #299, refs #215

### DIFF
--- a/pom-main.xml
+++ b/pom-main.xml
@@ -109,6 +109,11 @@
 							<className>com/esotericsoftware/reflectasm/**</className>
 							<differenceType>8001</differenceType>
 						</difference>
+						<difference>
+							<className>com/esotericsoftware/kryo/Kryo</className>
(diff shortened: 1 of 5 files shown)


johnynek:
:+1: we need this for https://github.com/twitter/chill/issues/252

magro:
@romix WDYT?

romix:
No objections from me.

PR #417, opened 2016-04-04
make FieldSerializer.serializeTransient usable

leewin12:
It is following pull request of #402.
I am sorry to make redundant pull request.
As you advise, I update against #414

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
@@ -98,9 +98,6 @@ public class FieldSerializer<T> extends Serializer<T> implements Comparator<Fiel
 	 * </p> */
 	private boolean useMemRegions = false;
 
-	/** If set, transient fields will be serialized */
-	private final boolean serializeTransient = false;
-
 	private boolean hasObjectFields = false;
 
 	static CachedFieldFactory asmFieldFactory;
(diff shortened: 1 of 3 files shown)


src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
```diff
@@ -602,6 +604,12 @@ public void removeField (CachedField removeField) {
 		return fields;
 	}
 
+    /** Get all transient fields controlled by this FieldSerializer
+     * @return all transient  fields controlled by this FieldSerializer */
+	public CachedField[] getTransientFields () {
```
magro: What is this needed for? We should keep the public api as small as possible...

leewin12: It is for custom serializer.
In case of custom serializer, there is no way to get transient fields so that it is hard to serialize transient fields.
I think that making `fields` and `transientFields` from private to protected could be good alternative if you decide to do not add more public api.

magro: For this use case you can also use `removeField(fieldName)`, which also handles transient fields. Is this sufficient then?

leewin12: Oops, that's true.
That's was temporary measure before I made a pull request of it.
Soon, I will get rid of the public getTransientFields.
Thank you for your advice.

leewin12:
I removed getTransientFields as you've advised.
Take a look and merge it please.

magro:
Thanks for the PR!

PR #418, opened 2016-04-10
Throw exception if output buffer size is greater than max buffer size

WilliamDo:
Currently, the Output class allows the initial buffer size to be greater than the max buffer size. In some cases, we can observe an ArrayIndexOutOfBoundsException during an attempt to expand the buffer. The following test cases demonstrate the behaviour.

```
    @Test
    public void testNewOutputMaxBufferSizeLessThanBufferSize() {
        int bufferSize = 2;
        int maxBufferSize = 1;

        Output output = new Output(bufferSize, maxBufferSize);

        for (int i = 0; i < bufferSize; i++) {
            output.writeByte(0);
        }

        output.writeByte(0);

    }

    @Test
    public void testSetOutputMaxBufferSizeLessThanBufferSize() {
        int bufferSize = 2;
        int maxBufferSize = 1;

        Output output = new Output(bufferSize, bufferSize);
        output.setBuffer(new byte[bufferSize], maxBufferSize);

        for (int i = 0; i < bufferSize; i++) {
            output.writeByte(0);
        }

        output.writeByte(0);

    }

```
Suggestion is to throw an IllegalArgumentException when setting the buffer sizes to pre-empt this error.

### DIFF
--- a/src/com/esotericsoftware/kryo/io/Output.java
+++ b/src/com/esotericsoftware/kryo/io/Output.java
@@ -54,6 +54,8 @@ public class Output extends OutputStream {
 	 * @param maxBufferSize The buffer is doubled as needed until it exceeds maxBufferSize and an exception is thrown. Can be -1
 	 *           for no maximum. */
 	public Output (int bufferSize, int maxBufferSize) {
+		if (bufferSize > maxBufferSize && maxBufferSize != -1)
+			throw new IllegalArgumentException("bufferSize: " + bufferSize + " cannot be greater than maxBufferSize: " + maxBufferSize);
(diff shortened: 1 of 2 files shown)


magro:
LGTM. Only the test might go into InputOutputTest.java, right?

WilliamDo:
Yes that makes sense. Moved the tests and as a result, also found my mistake with maxBufferSize=-1 which is valid case being less than the initial buffer size. I've also updated the code for this.

magro:
Cool. Can you squash the commits? Then I'll merge. Thanks!

WilliamDo:
No worries, all done.

magro:
Thanks!

PR #424, opened 2016-05-13
Fix for CompatibleFieldSerializer inheritance issue

xfrendx:
My try to fix the PR #187. This solution here doesn't ignore "duplicate" fields, because they are not duplicate at all. They just belongs to different classes... 

Added a central method FieldSerializer#getFieldName(CachedField) that returns per default the field name. This behaviour is overriden in CompatibleFieldSerialzer and can be driven by the "extendedFieldNaming" boolean. If extendedFieldNaming is true, then Class simple name is added to the field name... and this helps to find the right field when deserializing. It's not perfect, but don't want to have it too verbose.

Any suggestions will be welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,7 @@ A serialization library needs special knowledge on how to create new instances, 
 
 The Serializer class has a `copy` method that does the work. These methods can be ignored when implementing application specific serializers if the copying functionality will not be used. All serializers provided with Kryo support copying. Multiple references to the same object and circular references are handled by the framework automatically.
 
-Similar to the `read()` Serializer method, `kryo.reference()` must be called before Kryo can be used to copy child objects. See [Serializers](#Serializers) for more information.
+Similar to the `read()` Serializer method, `kryo.reference()` must be called before Kryo can be used to copy child objects. See [Serializers](#serializers) for more information.
 
 Similar to KryoSerializable, classes can implement KryoCopyable to do their own copying:
 
@@ -561,7 +561,20 @@ VersionFieldSerializer extends FieldSerializer and allows fields to have a `@Sin
 
 TaggedFieldSerializer extends FieldSerializer to only serialize fields that have a `@Tag(int)` annotation, providing backward compatibility so new fields can be added. And it also provides forward compatibility by `setIgnoreUnknownTags(true)`, thus any unknown field tags will be ignored. TaggedFieldSerializer has two advantages over VersionFieldSerializer: 1) fields can be renamed and 2) fields marked with the `@Deprecated` annotation will be ignored when reading old bytes and won't be written to new bytes. Deprecation effectively removes the field from serialization, though the field and `@Tag` annotation must remain in the class. Deprecated fields can optionally be made private and/or renamed so they don't clutter the class (eg, `ignored`, `ignored2`). For these reasons, TaggedFieldSerializer generally provides more flexibility for classes to evolve. The downside is that it has a small amount of additional overhead compared to VersionFieldSerializer (an additional varint per field).
 
-CompatibleFieldSerializer extends FieldSerializer to provide both forward and backward compatibility, meaning fields can be added or removed without invalidating previously serialized bytes. Changing the type of a field is not supported. Like FieldSerializer, it can serialize most classes without needing annotations. The forward and backward compatibility comes at a cost: the first time the class is encountered in the serialized bytes, a simple schema is written containing the field name strings. Also, during serialization and deserialization buffers are allocated to perform chunked encoding. This is what enables CompatibleFieldSerializer to skip bytes for fields it does not know about. When Kryo is configured to use references, there can be a [problem](https://github.com/EsotericSoftware/kryo/issues/286#issuecomment-74870545) with CompatibleFieldSerializer if a field is removed.
+CompatibleFieldSerializer extends FieldSerializer to provide both forward and backward compatibility, meaning fields can be added or removed without invalidating previously serialized bytes. Changing the type of a field is not supported. Like FieldSerializer, it can serialize most classes without needing annotations. The forward and backward compatibility comes at a cost: the first time the class is encountered in the serialized bytes, a simple schema is written containing the field name strings. Also, during serialization and deserialization buffers are allocated to perform chunked encoding. This is what enables CompatibleFieldSerializer to skip bytes for fields it does not know about. When Kryo is configured to use references, there can be a [problem](https://github.com/EsotericSoftware/kryo/issues/286#issuecomment-74870545) with CompatibleFieldSerializer if a field is removed. In case your class inheritance hierarchy contains same named fields, use the `CachedFieldNameStrategy.EXTENDED` strategy.
+
+```java
+class A {
+	String a;
+}
+
+class B extends A {
+	String a;
+}
+...
+// use `EXTENDED` name strategy, otherwise serialized object can't be deserialized correctly. Attention, `EXTENDED` strategy increases the serialized footprint.
+kryo.getFieldSerializerConfig().setCachedFieldNameStrategy(FieldSerializer.CachedFieldNameStrategy.EXTENDED);
+```
 
 Additional serializers can easily be developed for forward and backward compatibility, such as a serializer that uses an external, hand written schema.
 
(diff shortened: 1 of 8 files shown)


magro:
Intersting approach, great idea!

Suggestions:
* I'd prefer to have the test in the existing `CompatibleFieldSerializerTest`.
* The `extendedFieldNaming` config setting should go into a `CompatibleFieldSerializerConfig` which should also be added to `Kryo`, see also `TaggedFieldSerializerConfig`.
* Maybe we should also mention this setting in the README?

@NathanSweet WDYT?

xfrendx:
I did some changes:

* test now in the existing `CompatibleFieldSerializerTest`
* added `CachedFieldNameStrategy` with two implementations `DEFAULT` AND `EXTENDED`
* per default, the `DEFAULT` strategy is activated
* you can switch with `kryo.getFieldSerializerConfig().setCachedFieldNameStrategy(FieldSerializer.CachedFieldNameStrategy.EXTENDED);`
* README will be extended after this solutions passes the kryo technical QA

magro:
I wonder if it makes any sense for `FieldSerializer` to use anything else than default - perhaps only allow to override `getFieldName` and move/keep the strategy close to the `CompatibleFieldSerializer` in its own config. WDYT?

xfrendx:
There are some methods in `FieldSerializer` that uses the CacheField name:
* `getField`
* `removeField`
* `compare`

... all this methods would need to be overriden in `CompatibleFieldSerializer`,  otherwise they won't work correctly (evt. you can be confused by `getField`...what `CachedField` you get when multiple same named fields are present). 

... but I got your point. I don't like to attack the `FieldSerializer` code

xfrendx:
Hi @magro,

would like to go ahead with this PR, but I'm confused somehow. The getField(String) and removeField(String) in FieldSerializer can't work correctly with the current/default approach. They get/remove only the first field with same name... but if there are more fields with same name (but different class), they give wrong result.

I can do now the fix only for CompatibleFieldSerializer...but need then override the getField/removeField methods.... To fix it in FieldSerializer already seems to be much better for me.

WDYT?

magro:
Sounds reasonable to fix FieldSerializer, can you do this in a separate commit or even separate PR? Adding a test for this as well would be super awesome :-)

xfrendx:
Hi @magro,

the new PR would have identical commits like this one ;) If you think it's neccessary, I'll do it. But pls. explain me how it should help us?

I'll try to sum up the new features/bugfixes that are comming with this PR:
* added `CachedFieldNameStrategy` that can be set to `FieldSerializerConfig`
* strategy influences how `CachedFields` are recognized in:
 * `FieldSerializer.getField(String)`
 * `FieldSerializer.removeField(String)`
 * `FieldSerializer.compare(CachedField, CachedField)`
 * `CompatibleFieldSerializer.read(...)`
* `DEFAULT` strategy uses just field name
* `EXTENDED` strategy uses Class simple name and field name. That helps in better identification in case that there are same named fields in class inheritance path.
 * E.g. you can't get the right field via `FieldSerializer.getField` method without
 * `CompatibleFieldSerializer` doesn't work correctly...you'll not having the same data back after storing to Kryo

magro:
A PR with a bugfix and a test usually can be merged fairly fast. Therefore I mentioned this option. You can also add the commit to this PR. Please add a test that shows the issue for FieldSerializer as well. Thanks!

magro:
@xfrendx We're going to have a release soon, what do you think when you can finalize this?

xfrendx:
Hi @magro,

I added the TCs right now, pls. check it. Thx

test/com/esotericsoftware/kryo/FieldSerializerTest.java
```diff
@@ -621,7 +621,71 @@ public void testMultipleTimesAnnotatedMapFields () {
 
 		assertFalse("Exception was expected", true);
 	}
-	
+
```
magro: If you don't mind you could put this into a `FieldSerializerInheritanceTest` - sorry that I've suggested to add this to this test, I wasn't aware that it's already that big...

test/com/esotericsoftware/kryo/FieldSerializerTest.java
```diff
@@ -621,7 +621,71 @@ public void testMultipleTimesAnnotatedMapFields () {
...
+		assertFieldRemoved(serializer, "TestDefault.a");
+		assertNotNull(serializer.getField("TestExtended.a"));
+		serializer.removeField("TestExtended.a");
+		assertFieldRemoved(serializer, "TestExtended.a");
+	}
+
+	protected void assertFieldRemoved(FieldSerializer serializer, String fieldName) {
+		try {
+			assertNull(serializer.getField(fieldName));
+			assertTrue(false);
```
magro: Could be `Assert.fail("Expected IllegalArgumentException to be thrown for serializer.getField(" + fieldName + ")")`.

src/com/esotericsoftware/kryo/serializers/FieldSerializerConfig.java
```diff
@@ -136,4 +138,12 @@ public boolean isCopyTransient() {
     public boolean isSerializeTransient() {
         return serializeTransient;
     }
+
+    public FieldSerializer.CachedFieldNameStrategy getCachedFieldNameStrategy() {
+        return cachedFieldNameStrategy;
+    }
+
+    public void setCachedFieldNameStrategy(FieldSerializer.CachedFieldNameStrategy cachedFieldNameStrategy) {
+        this.cachedFieldNameStrategy = cachedFieldNameStrategy;
```
magro: Add `if (TRACE) trace("kryo.FieldSerializerConfig", "CachedFieldNameStrategy: " + cachedFieldNameStrategy);` as in other methods.

README.md
```diff
@@ -561,7 +561,20 @@ VersionFieldSerializer extends FieldSerializer and allows fields to have a `@Sin
...
+CompatibleFieldSerializer extends FieldSerializer to provide both forward and backward compatibility, meaning fields can be added or removed without invalidating previously serialized bytes. Changing the type of a field is not supported. Like FieldSerializer, it can serialize most classes without needing annotations. The forward and backward compatibility comes at a cost: the first time the class is encountered in the serialized bytes, a simple schema is written containing the field name strings. Also, during serialization and deserialization buffers are allocated to perform chunked encoding. This is what enables CompatibleFieldSerializer to skip bytes for fields it does not know about. When Kryo is configured to use references, there can be a [problem](https://github.com/EsotericSoftware/kryo/issues/286#issuecomment-74870545) with CompatibleFieldSerializer if a field is removed. In case your class inheritance hierarchy contains same named fields, use the `CachedFieldNameStrategy.EXTENDED` strategy.
+
+```java
+class A {
+	String a;
+}
+
+class B extends A{
```
magro: Nitpick: space missing after `A` :-)

magro:
@xfrendx Thanks, I added minor/cosmetic comments. Once they're resolved please squash the changes to a single commit and change the commit message to mention both FieldSerializer and CompatibleFieldSerializer. I'll then merge the PR.

xfrendx:
@magro your suggestions has been added and changes suqashed to one commit. Let me know if there is something left...

test/com/esotericsoftware/kryo/FieldSerializerInheritanceTest.java
```diff
@@ -0,0 +1,129 @@
+package com.esotericsoftware.kryo;
```
magro: Please add the license header.

magro:
@xfrendx Great, thanks! In the new test only the license header is missing. Please change the commit message to s.th. like "Support 'duplicate'/overwritten fields in (Compatible)FieldSerializer" and add a description with the [what and why](http://chris.beams.io/posts/git-commit/#why-not-how) of the change. This would be perfect :-)

xfrendx:
@magro done

magro:
@xfrendx Following general rules on commit messages they should have a short title/subject identifying the change. The description should be separated from the title with an empty line. See also http://chris.beams.io/posts/git-commit/

xfrendx:
@magro thx for you patience & done

magro:
@xfrendx Awesome, thanks! :-)

PR #429, opened 2016-06-03
Fix debug message when overwriting registration

### LINKED ISSUE #421: Misleading debug message when registering classes
To reproduce:

```
Log.DEBUG();
kryo = new Kryo();
kryo.register(SomeObject.class, new SomeObjectSerializer(), 1);
```
Resulting log:
```
DEBUG: An existing registration with a different type already uses ID: 1
Existing registration: [1, String]
Unable to set registration: [1, com.mypackage.SomeObject]
```

The Javadoc on the other hand says that existing registrations are merely replaced. And from briefly viewing the code, it appears to proceed with registration.

cypherdare:
Fixes #421


magro:
Thanks!

Issue #411, opened 2016-03-18
Kryo fail to wipe out generic type for collection serializer when reference objects


yumengzhao92:
[kryo-bugreport.zip](https://github.com/EsotericSoftware/kryo/files/180437/kryo-bugreport.zip)

Blow is a brief description of when the bug happens(Please run KryoSerializationIssue in attached file to see what I mean): 

Given classes A, B
Class A {List<String> s1, List<String> s2}
Class B is an ArrayList<Double>

An instance a of A and an instance b of B are to be deserialized by Kryo.

What’s special about a and b and how they are combined to produce the bug is that:
- a.s2 is a reference of a.s1, and s2 is deserialized after s1
- b is deserialized after a
- b has the same class type as a.s2 (both of them are ArrayList in this case) but they have different generic types - the generic type of a.s2 is String, the generic type of b is Double

When a.s2 is to be deserialized, the CollectionSerializer’ genericType in ArrayList Registration is set to String. This genericType is usually wiped out (set to null) when CollectionSerializer is entered to read elements of the collection.
But this step is bypassed because a.s2 is a reference to earlier object and can be retrieved from recorded objects thus CollectionSerializer is not entered.
This has caused the CollectionSerializer of ArrayList Registration inside classToRegistration Map to hang on to the uncleared generic type String, even if the next ArrayList to be deserialized is a Collection of Doubles (or any other non-primitive types).

When deserialize b, Kryo is trying to deserialize String instead of Double and cause problems.

So to generalize the case, i believe an internal bug exists when Kryo deserializes two classes in such a way that,
the last field to be deserialized of the first class is the same collection type as the second class, and the last field instance is a reference to an earlier object.
This will cause Kryo to carry over the wrong generic type to deserialize the second class and throw errors.

yumengzhao92:
One possible solution might be to add 
`serializer.setGenerics(this, null)`
to Kryo.readObject and Kryo.readObjectOrNull method if (stackSize == REF)

```
** Reads an object using the specified serializer. The registered serializer is ignored. */
	public <T> T readObject (Input input, Class<T> type, Serializer serializer) {
		if (input == null) throw new IllegalArgumentException("input cannot be null.");
		if (type == null) throw new IllegalArgumentException("type cannot be null.");
		if (serializer == null) throw new IllegalArgumentException("serializer cannot be null.");
		beginObject();
		try {
			T object;
			if (references) {
				int stackSize = readReferenceOrNull(input, type, false);
[...]

PR #433, opened 2016-06-11
Provide an opt-out for disabling the optimization of generics serialization

romix:
The existing optimization of generics serialization aims at reducing the size of the serialized representation. This optimization is enabled by default. But sometimes it may have a negative impact on performance. And sometimes it cannot handle complex cases of generics usage.

This patch changes the default behavior and disables this this optimization. To enable it, simply invoke `FieldSerializerConfig.setOptimizedGenerics(true)`.


magro:
AFAICS the following issues might benefit from this PR and should be checked with the changes of this PR: #377, #384 (#385), #411

src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
```diff
@@ -331,10 +340,10 @@ protected void initializeCachedFields () {
 
 	CachedField newCachedField (Field field, int fieldIndex, int accessIndex) {
 		Class[] fieldClass = new Class[] {field.getType()};
-		Type fieldGenericType = field.getGenericType();
+		Type fieldGenericType = (config.isOptimizedGenerics()) ? field.getGenericType() : null;
 		CachedField cachedField;
 
-		if (fieldGenericType == fieldClass[0]) {
+		if (!config.isOptimizedGenerics() || fieldGenericType == fieldClass[0]) {
 			// This is a field without generic type parameters
```
magro: This comment should probably be changed now.

src/com/esotericsoftware/kryo/serializers/FieldSerializerConfig.java
```diff
@@ -37,6 +37,8 @@
     private boolean copyTransient = true;
     /** If set, transient fields will be serialized */
     private boolean serializeTransient = false;
+    /** Try to optimize handling of generics */
```
magro: add s.th. like "[..] for smaller size"?

test/com/esotericsoftware/kryo/FieldSerializerTest.java
```diff
@@ -392,6 +392,7 @@ public void testStdInstantiatorStrategy () {
 	
 	public void testGenericTypes () {
 		kryo = new Kryo();
+		kryo.getFieldSerializerConfig().setOptimizedGenerics(true);
```
magro: duplicate this test with `setOptimizedGenerics(false);`?

test/com/esotericsoftware/kryo/SerializationCompatTest.java
```diff
@@ -109,9 +109,24 @@ public void testDefaultSerializers() throws Exception {
                 EXPECTED_DEFAULT_SERIALIZER_COUNT, defaultSerializers.size());
     }
 
-    public void testStandard () throws Exception {
+    public void testStandardOptGenerics () throws Exception {
+   	  kryo.getFieldSerializerConfig().setOptimizedGenerics(true);
         runTests(
-                "standard",
+                "standard-opt-generics",
```
magro: The existing versioned files `test/resources/*-standard.ser` should be `git mv`'ed to `test/resources/*-standard-opt-generics.ser`.

romix:
@magro I'm not sure about the interplay of the config field of the FieldSerializer and my patch. If someone would change the optimizedGenerics setting of the config from `true` to `false` after a FieldSerializer was build using this config, then it could lead to undefined behavior. 

May be FieldSerializer should get a non-modifiable copy of the config?

test/com/esotericsoftware/kryo/SerializationCompatTest.java
```diff
@@ -109,9 +109,24 @@ public void testDefaultSerializers() throws Exception {
...
+                },
+                new Function1<File, Output>() {
+                    public Output apply(File file) throws Exception {
+                        return new Output(new FileOutputStream(file));
+                    }
+                }
+        );
+        kryo.getFieldSerializerConfig().setOptimizedGenerics(false);
+        runTests(
+                "standard-nonopt-generics",
```
magro: The new files (`test/resources/*-standard-nonopt-generics.ser`) created when running this `SerializationCompatTest` should be added to git.

test/com/esotericsoftware/kryo/SerializationCompatTest.java
```diff
@@ -109,9 +109,24 @@ public void testDefaultSerializers() throws Exception {
...
+                        return new Input(new FileInputStream(file));
+                    }
+                },
+                new Function1<File, Output>() {
+                    public Output apply(File file) throws Exception {
+                        return new Output(new FileOutputStream(file));
+                    }
+                }
+        );
+        kryo.getFieldSerializerConfig().setOptimizedGenerics(false);
```
magro: This would not affect the config of existing, previously created `FieldSerializer` instances, because they use `kryo.getFieldSerializerConfig().clone()`. Either we setup the kryo instance again with a different opt-gen setting or reconfigure existing `FieldSerializer` instances.

magro:
@romix Thanks again for your work here! Can you elaborate a bit more on the consequences of the `optimizedGenerics` setting being true or false (e.g. with some examples and size/time measurements)? If you don't have the time for this I could also help with building a simple benchmark.

As I understand it, with non-optimized generics we should have faster (de)serialization of generics but increased size, and generics serialization should be more robust. Assuming that generics serialization is in-fact more robust with this setting, I wonder if we should make this the default (if we'd release this then as 4.0 or 3.1 is a different question). WDYT?

/cc @NathanSweet

magro:
@romix The FieldSerializer could also "freeze" this setting in the cloned
config by setting a special flag, so that the setter throws a exception if
called on the clone.

romix <notifications@github.com> schrieb am Mo., 13. Juni 2016, 00:18:

romix:
@magro Regarding the consequences of this patch on size/performance: I haven't written or executed any benchmarks due to the lack of time. But your understanding is correct. I expect exactly the same as you do: with non-optimized generics we should have faster (de)serialization of generics but increased size, and generics serialization should be more robust. Basically, the size is likely the be roughly the same as it used to be before I introduced those generics optimizations a while ago. 

It would be nice if you or someone else could help with the benchmarking of this non-optimized mode.

As for making it a default, I'm not sure. But it could be a good idea, indeed. Though, it would be a breaking change, because it changes the default binary serialization format in case of generics. People willing to proceed with the existing format will have to change their source code and enable the setting by hand.

magro:
@romix Sorry for the late response, this week was very busy.

I'm going to submit a PR with my comments / proposed changes against your branch. If you're already working on it please ping me now so that we don't do double work.

Not sure if I'll have time for the benchmark as well, perhaps I'll submit a separate PR later against master. 

Making the non-optimized mode the default: this is my preference, as it should improve stability and reduce submitted issues. I'll ask on the mailing list for opinions, let's see if there's feedback.

magro:
@romix I just submitted my feedback as [PR](https://github.com/romix/kryo/pull/3). If you think all this is fine you could merge + rebase/squash commits to a single one - and IMO we could merge it then. The most critical thing is that change for the optimizedGenerics default value :-)

romix:
@magro Thanks a lot for your PR!

romix:
@magro I force-pushed the new version of the patch. It includes all of your changes and my changes to avoid any hacks related to freezing of the `FieldSerializerConfig.setOptimizedGenerics`.

Please check if everything is OK and we could merge this.

magro:
@romix Sorry for the late response... Because optimization is off by default the commit message is wrong in this regard. Can you update it? The readme probably should also mention this setting, but I could write that later as well.

romix:
@magro Good catch! I updated the commit message. The readme could be updated later, when we roll out a new release.

magro:
So it's done now, great work!!

romix:
@magro Martin, I'm not sure that closing all those related issues is the right thing to do. I agree that with #433 they are not present in the non-generic-optimized mode. But they are still present when generic optimizations are enabled. Therefore, I'd suggest tagging them with "optimize-generics" or something like this, instead of closing. WDYT?

magro:
@romix Sounds good!

Issue #291, opened 2015-03-10
Improve field removal in CompatibleFieldSerializer


solf:
This is an extension of https://github.com/EsotericSoftware/kryo/issues/286

Please improve removed field handling in CompatibleFieldSerializer so that it is actually possible to remove fields (rather than having to keep them indefinitely for backward compatibility).

Consider the patch I've come up with: https://copy.com/akbTt9twsSr2CNqU
And second patch to force exception on missing fields (avoids risk on incorrect de-serialization; must be applied after patch above): https://copy.com/hk1YcLDhxqfMaylJ

This does two things:
- Obsolete fields can be removed into inner(nested) class (ObsoleteKryoFields) that ought to be static + private. This way these obsolete fields don't consume any memory when instantiating original class and don't clutter its code (other than the inner class itself).
- If there are obsolete fields during de-serialization, then if class declares 'private transient ObsoleteKryoFields obsoleteKryoFields' field, then its value will be set to the instance holding obsolete fields (thus an easy support for migrating obsolete values into their new 'homes'). This also works across superclasses (hence the requirement for the field to be 'private').

If there are no actual obsolete fields in the data being de-serialized, then the performance hit is minimal.

For testing/demonstration:
- Start with http://pastebin.com/e8RpFkfm -- create serialized data on disk.
- Test with something like this: http://pastebin.com/9eWD0vha You can also remove/comment out ObsoleteKryoFields inner class(es) and watch how de-serialization fails when fields are removed without this additional support.

NathanSweet:
Can you submit the patch as a PR so it is more easily considered and merged?

solf:
Not easily, no. I'm not familiar with the term & what it entails.

The diff I've provided is pretty trivial (basically CompatibleFieldSerializer is changed + one field in CachedField). If this is truly a problem for some reason, I'll see if I can do PR (whatever that means).

Oh, and I think I forgot to mention -- this is patch for 2.4.0 release.

magro:
PR stands for pull request, you can read about this here:
https://help.github.com/articles/creating-a-pull-request/

Cheers,
Martin
Am 11.03.2015 13:33 schrieb "solf" <notifications@github.com>:

solf:
Well, I tried to create PR today, but I only have very limited GIT knowledge and trying to guess what I'm supposed to do, this is what I ended up with trying to push the branch I've created locally:
[...]

PR #437, opened 2016-06-28
FieldMappingCompatableSerializer to patch CompatibleFieldSerializer and It's UnitTest

diaimm:
FieldMappingCompatableSerializer to patch CompatibleFieldSerializer 
and It's UnitTest

by diaimm


magro:
I'm missing context, what's the value of this PR, the new serializer? If it was discussed previously somewhere please add a link.

magro:
While going through open issues I found the related issue: #291. As reported there there still seems to be an issue with this change.

NathanSweet:
OP hasn't followed up.

Issue #440, opened 2016-07-03
TaggedFieldSerializer tag values cannot overlap those of superclass


cypherdare:
It's not immediately clear that you cannot tag a field with the same number as a tagged field in a superclass. And even if one is aware of this requirement, we must be very vigilant about checking class hierarchies for the possible clashes. I'm suggesting that during serialization, TaggedFieldSerializer could check for tag value clashes in the hierarchy and throw an error if any exist.

On a related note, it might also be nice if superclasses are checked for complete omission of tags and a warning shown if they don't have any.

magro:
Sounds good! For the second case perhaps one might also want to force an exception, not sure. Do you want to submit a pull request (or 2)?
Btw, what happens in the first case?

cypherdare:
I'll give it a shot when I have some free time. It looks like for the first issue, we don't need to look at class hierarchy...we only need to check for any reused tag values since that's all the problem really is.

During serialization, TaggedFieldSerializer makes no complaints. I haven't dug into what is actually happening, but in my case, I got a ClassNotFoundException during deserialization because a class name was corrupted.

For the second case, I don't know if an exception is right because there are possible (rare) cases where you might want to omit all the fields of a superclass.

magro:
Ok, great. Regarding the second case I meant that we add a switch to choose between warning and exception.

cypherdare:
I have an idea for how to do the second case. But I'm not sure if the warning should only be shown at the TRACE or DEBUG log level. Is there already a precedent for warnings like this?

I suppose the option to throw an error could go into TaggedFieldSerializerConfig, if we need that. However, I'm thinking that if someone knows enough about this issue to turn the setting on, they'd already know to be checking the logs during testing.

magro:
Perhaps we can allow to choose from throw exception, warn and debug?

NathanSweet:
Less configuration is always better, if it can be avoided. If we aren't sure, avoid it.

I'm not sure the warning makes sense though. It could be fine that a super class has no tags, otherwise it is likely it would be discovered quickly. If you'd like to add it, I'd choose TRACE. I'll close this issue, feel free to submit a PR.

PR #441, opened 2016-07-08
Check for overlapping tags

cypherdare:
Addresses #440.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/TaggedFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/TaggedFieldSerializer.java
@@ -42,6 +42,9 @@ import com.esotericsoftware.kryo.io.Output;
  * don't clutter the class (eg, <code>ignored</code>, <code>ignored2</code>). For these reasons, TaggedFieldSerializer generally
  * provides more flexibility for classes to evolve. The downside is that it has a small amount of additional overhead compared to
  * VersionFieldSerializer (an additional varint per field). Forward compatibility is not supported.
+ * <p>
+ *     Tag values must be entirely unique, even among a class and its superclass(es). An IllegalArgumentException will be
+ *     thrown by {@link Kryo#register(Class)} (and its overloads) if duplicate Tag values are encountered.
  * @see VersionFieldSerializer
  * @author Nathan Sweet <misc@n4te.com> */
 public class TaggedFieldSerializer<T> extends FieldSerializer<T> {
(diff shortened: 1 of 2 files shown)


magro:
Thanks for the PR, and sorry for the delay (I was on vacation with bad connectivity). 

Can you add a test for the added check?

cypherdare:
@magro Sorry to take so long. I got burned by this issue again and spent three hours before discovering it. So I remembered that I had this PR waiting for clean-up. Should be ready now.

magro:
Great, thanks!

magro:
@cypherdare One follow up question: what's the recommendation for users that get this exception for their existing code base after the kryo update (assuming the underlying issue was not noticed before)? We should at least tell this in the release notes, maybe we could add some actionable hint to the exception.

cypherdare:
@magro 
So, thinking through ways in which someone may have released something with overlapping tags without noticing the error:

1. The overlapping tags were for fields that use the same number of bytes when written. Reading the data is safe, but one or both of the two fields has the incorrect value in it, so it is conceivable that this might not have been caught in testing. If the read order and write order are the same, then the second of the two fields will at least be correct. The order might have changed if field names have changed.
2. The overlapping tags were for the last two fields written, and the object doesn't appear as part of a larger object graph (because if it did, subsequent data in the stream would be read from the wrong location and an error would have occurred). In this case, the data in both fields is probably corrupted.
3. There is an extremely small chance that something could have made it through testing without either of the above occurring.

We might be able to conceivably recover data if one of the above two is true. If the field names have not been changed, the two fields were *probably* written in alphabetical order, so when reading them, we could map multiple appearances of the same Tag ID sequentially to their original fields. Maybe someone could pass in a Map<Integer, List<Integer>> linking old tag values to new tag values as a configuration. This is an awful lot of complexity to add to the TaggedFieldSerializer class for a small possibility. Perhaps a RecoveryTaggedFieldSerializer class could be written and put in a Gist for those that might want to attempt this recovery. But I don't know that it's worth trying to adequately test a solution like this if we don't know if anyone has been affected.

For situation 1, if we can assume the fields were written in alphabetical order*, the simpler solution is to give one of the two fields a new unique tag value, and the other should be left alone. The one that keeps the tag value should be the one that is last alphabetically. During reading, it will be written once with the wrong value, and then a second time with the correct value. The class should be updated to be able to withstand the first field's data being lost.

*I'm not 100% this can be assumed. It looks to me like fields are naturally in alphabetical order. We started sorting them by tag ID at some point, but fields with the same ID *probably* didn't switch places during sorting. If we can't assume this, we still need to keep that tag on one of the values to prevent errors during reading (like losing our place in the stream), but the class must be updated to assume that both of the fields might have incorrect values.

For situation 2, I think both fields need new IDs, and the original tag ID should be deprecated, which can be done by creating some new unused field and tagging it with that ID and deprecating it. Old data should still be readable, and the application can now safely write the data as part of a larger graph.

For situation 3, I think the same should be done as for situation 2, but there should also be error checking on the object graph after reading it if attempting to read old data.

magro:
@cypherdare Wow, thanks for this elaborate analysis! When preparing the release notes I'll add a link to your comment here, ok?

cypherdare:
@magro No problem. There could be errors here but if someone encounters this issue, we can probably reason out a solution.

PR #450, opened 2016-08-01
updating the bcel version since the 6.0-SNAPSHOT no longer resolves

lbergelson:
as discussed on gitter, this fixes a build failure

### DIFF
--- a/pom-main.xml
+++ b/pom-main.xml
@@ -122,7 +122,7 @@
 					<dependency>
 						<groupId>org.apache.bcel</groupId>
 						<artifactId>bcel</artifactId>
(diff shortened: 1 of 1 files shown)


magro:
Thanks for the PR. There might also be a new version of the clirr plugin available with the updated dependency, perhaps we can use this soon. See https://github.com/mojohaus/clirr-maven-plugin/issues/19

lbergelson:
I only see the 2.7 version available which still depends on bcel 5.2.  Happy to update my pull request when 2.8 is in maven central.

magro:
The clirr maven plugin 2.8 was just released btw.

Issue #462, opened 2016-09-27
StackOverflowError on 2.24.0 and 3.0.3 when serializing tree-like object


marianoguerra:
hi, I'm getting a StackOverflowError on kryo 2.24.0 (the one that flink uses), I tried it in 3.0.3 and it also happens, I made a small project that reproduces the problem, it's here:

https://github.com/marianoguerra-atik/kryo-serialize-tree-stack-overflow

the project basically has an interface NodeBase<T> and two classes that implement it:
- `Node<T>`
- `Leaf<T>`

`Leaf<T>` holds a value of T
`Node<T>` holds a `Map<String, NodeBase<T>>`

this builds a tree-like structure, that when serialized fails with the error attached at the end of the issue.

you can reproduce it by running the unit test on the 3 branches:

```
# fails
git checkout kryo-2.24.0
mvn clean test

# fails
git checkout kryo-3.0.3
mvn clean test

# passes
git checkout kryo-4.0.0
mvn clean test
```

The test is simple:

```
Kryo kryo = new Kryo();

Output output = new Output(new FileOutputStream("file.bin"));

Node<Long> node = new Node<Long>();
Node<Long> c1 = new Node<Long>();
node.set("c1", c1);
c1.set("c2", new Leaf<Long>(42L));

kryo.writeObject(output, node);
output.close();

Input input = new Input(new FileInputStream("file.bin"));
Node<Long> result = kryo.readObject(input, Node.class);
input.close();
```

Is it possible to backport whatever fixes the problem from 4.0.0 to 2.24.1 and 3.0.4?

if not, is it something I can do in my code to avoid this problem?

Thanks in advance

The error is:

```
T E S T S
-------------------------------------------------------
Running com.gnandoo.AppTest
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.125 sec <<< FAILURE! - in com.gnandoo.AppTest
testApp(com.gnandoo.AppTest)  Time elapsed: 0.123 sec  <<< ERROR!
com.esotericsoftware.kryo.KryoException: java.lang.IllegalArgumentException: Unable to create serializer "com.esotericsoftware.kryo.serializers.FieldSerializer" for class: com.gnandoo.Leaf
Serialization trace:
childs (com.gnandoo.Node)
childs (com.gnandoo.Node)
    at com.esotericsoftware.kryo.serializers.ObjectField.write(ObjectField.java:82)
[...]

PR #463, opened 2016-09-27
don't set parent of genericsScope if equals (#462)

marianoguerra:
this avoids a StackOverflowError when calling kryo.Generics.getConcreteClass

NOTE: this PR must be merged on top of kryo-2.24.0, the same patch should be applied on top of 3.0.3

I can also update the pom to 2.24.1 in this PR if needed


magro:
According to #462 this can be closed.

Issue #464, opened 2016-10-06
Wrong byte order if the ByteBuffer of the ByteBufferOutput has to grow during writeVarInt / writeVarLong


ahilsend:
Growing the ByteBuffer while writing varInt or varLong will change the byte order first, and then perform the require.

If the ByteBuffer needs to grow the setBuffer will again change the ByteOrder, but this time the one set by the varInt method. Thus the final ByteOrder was incorrect.

magro:
Closed with #465

PR #465, opened 2016-10-06
Fix growing ByteBufferOutput while writing varint (#464)

ahilsend:
Growing the ByteBuffer while writing varInt or varLong will change the byte order first, and then perform the require.

If the ByteBuffer needs to grow the setBuffer will again change the ByteOrder, but this time the one set by the varInt method. Thus the final ByteOrder was incorrect. (see [#464](https://github.com/EsotericSoftware/kryo/issues/464))


romix:
Looks good to me. Thanks!

PR #480, opened 2016-12-01
Adding marklet generation profile into POM and a first sample generation

Faylixe:
Adding an up-to-date generated Javadoc in Markdown format ready to use directly on Github using Marklet project. Looking for complex project like Kryo that can lead this docklet to a must have javadoc generator. Would not be mad if refused, but i would love to have a feedback about why you won't use it or why you would, and the kind of additional features you would require.

Regard ;)


NathanSweet:
It's interesting, but I'd rather not need to keep the markdown up to date.

PR #483, opened 2016-12-15
Override ObjectInputStream ClassLoader

aviemzur:
Due to a known Java issue ObjectInputStream might use a different ClassLoader
This can cause ClassNotFoundException


aviemzur:
R: @magro 
I encountered this issue in `Spark` in yarn cluster mode.
Spark executors use a specific `ClassLoader`, and using `JavaSerializer` I got `ClassNotFoundException`.
This change solves this making sure Kryo's `ClassLoader` is used when serializing.

magro:
Thanks for the PR!

PR #484, opened 2016-12-19
Remove graphContext usage from JavaSerializer

aviemzur:


aviemzur:
R: @magro 
Graph context lookups in `JavaSerializer` always return `null` which leaves the lookup redundant.
See test which demonstrates this in the following branch: https://github.com/aviemzur/kryo/commit/7f384f4eb5915fbd8e8de1b4e049ed49cd41d870

magro:
Not sure about the intended usage of GraphContext, but I can imagine there are ways to use so that it would not be null. Perhaps @NathanSweet or @romix can shed some light on this?

NathanSweet:
Likely the test is flawed. If JavaSerializer is used twice for a single object graph, then of course the second time the ObjectOutputStream from the first time is found in the graph context.

PR #486, opened 2017-01-04
Fix IdentityMap constructor

andilem:
Fix read of uninitialized variable in constructor of IdentityMap


magro:
Thanks for the PR!

Issue #430, opened 2016-06-03
Unclear that indices 0-9 already registered for primitives


cypherdare:
There is no way short of reading the source code that I've found for a user to know that Kryo has already used indices 0-9 to register all the primitives and String, and so can accidentally overwrite those registrations if explicitly setting the ID with `kryo.register(class, id)`. A typical user (if I'm one) starts registering classes at index 0 or 1 and up from there, thereby wiping out the registrations of all the primitives.

I'm not sure if there's a clean solution. It might be adequate to update the documentation to recommend assigning with indices >= 10. But when registration is set to required, and in light of #398, enforcement might be necessary.

magro:
Sounds reasonable.

@NathanSweet WDYT?

magro:
@NathanSweet WDYT about this?

magro:
Closed with #509

NathanSweet:
Sorry I'm late. Updating the docs in #509 is good.

Note that a user can clear and re-register all classes to have control of the ordinals.

PR #509, opened 2017-04-24
Clarify default used registrations

cypherdare:
Addresses #430 

The readme originally led me down the path of starting my registrations at 0 and wiping out the primitive registrations.

I suppose we could also log a warning if the ID is between 0 and 9.


magro:
Awesome, thanks!

Issue #511, opened 2017-04-27
Inet4Address round trip returns 0.0.0.0


jangoolie:
```scala
import com.esotericsoftware.kryo._
import java.net.Inet4Address
import io.{ Output, Input }
import java.io._

val kryo = new Kryo()

val output = new Output(new FileOutputStream("file.bin"))
val ip = java.net.InetAddress.getByName("123.123.123.123")
kryo.writeObject(output, ip)
output.close

val input = new Input(new FileInputStream("file.bin"));
val ip2 = kryo.readObject(input, classOf[Inet4Address])

scala> ip
res3: java.net.InetAddress = /123.123.123.123

scala> ip2
res4: java.net.Inet4Address = /0.0.0.0
```

:(

gurubook:
Same problem here, for Inet4Address and Inet6Address

`
package kryo.test;

import static org.junit.Assert.assertTrue;

import java.io.ByteArrayOutputStream;
import java.net.InetAddress;
import java.net.UnknownHostException;

import org.junit.Test;

import com.esotericsoftware.kryo.Kryo;
import com.esotericsoftware.kryo.io.Input;
import com.esotericsoftware.kryo.io.Output;

public class KryoSerializerTest {

	public static class Addresses {
		public Addresses() {
			super();	
		}
		
		public Addresses(InetAddress ipv4, InetAddress ipv6) {
			super();
			this.ipv4 = ipv4;
			this.ipv6 = ipv6;
		}
		
		private InetAddress ipv4;
		private InetAddress ipv6;
		
		public InetAddress getIpv4() {
			return ipv4;
		}

		public void setIpv4(InetAddress ipv4) {
			this.ipv4 = ipv4;
		}

		public InetAddress getIpv6() {
			return ipv6;
		}

		public void setIpv6(InetAddress ipv6) {
			this.ipv6 = ipv6;
		}
	}
	
	@Test
	public void inetAddressSerializationTest() throws UnknownHostException {
		Kryo kryo = new Kryo();
		
		// initialize addresses
		Addresses addresses = new Addresses();
		addresses.setIpv4(InetAddress.getByName("192.168.0.1"));
		addresses.setIpv6(InetAddress.getByName("::1"));
		assertTrue(addresses.ipv4.getHostAddress().equals("192.168.0.1"));
		assertTrue(addresses.ipv6.getHostAddress().equals("0:0:0:0:0:0:0:1"));
		
[...]

PR #513, opened 2017-05-26
Added serializers for Inet4Address and Inet6Address #511

jangoolie:
* Added Inet4/6Address serializers to DefaultSerializers
* Added tests for Inet4/6Address serializers

This relates to Issue https://github.com/EsotericSoftware/kryo/issues/511

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -1,15 +1,15 @@
 /* Copyright (c) 2008, Nathan Sweet
  * All rights reserved.
- * 
+ *
  * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
  * conditions are met:
- * 
+ *
  * - Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
  * - Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
  * disclaimer in the documentation and/or other materials provided with the distribution.
  * - Neither the name of Esoteric Software nor the names of its contributors may be used to endorse or promote products derived
  * from this software without specific prior written permission.
- * 
+ *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
  * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
  * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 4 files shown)


isaki-x:
I feel like these serializers should be opt in and part of a separate package (such as the UUID serializer). Serializing such data is not common in most applications.

NathanSweet:
It doesn't hurt much to have default serializers, it's one more entry in the list.

isaki-x:
I'm not trying to be pedantic, but why bother having another serializer library? Why not just make everything default?

Personally, I like the opt-in approach (especially since I have my own custom serialization and representation of IPAddress objects as well as my own Guava and UUID serializers tailored to my needs); it gives developers better insight (especially with white listed objects) what will and will not usable.

Ultimately InetAddress isn't dangerous from both a security and a performance perspective (at least in Java 6, 7, and 8) as long as the default implementations never call any ByName methods nor getHostName (as these invoke DNS; I don't want a DNS dependency in a serialization framework; especially once that I have chosen for its performance). I'm merely arguing my point of view on the matter. I haven't looked at the Java 9 implementation but I can't imagine it is wildly different than the current versions (address based creation doesn't call DNS until getHostName() is invoked).

isaki-x:
You know, there is another problem with default serialization of InetAddress classes; this implementation does what I think is probably the 90% case which is to transmit IP addresses between Kryo components (or store them for later retrieval). However, the InetAddress object stores more than just an IPAddress and this implementation does not serialize the other data (namely, the host name associated with the object if known).

The big problem is that you can't know whether or not DNS has been utilized or not without using reflection to access implementation specific private data as invoking any "get*Host*" methods will result in DNS being invoked if it hasn't already to fill in the missing data before returning an answer.

This probably doesn't seem too terrible until you start thinking about possible use cases. Let's say your focus is security or data auditing; perhaps you are recording what a DNS server returns for a particular cname over time.

It might seem natural to use InetAddress which, if created by getByName or getAllByName, already has the two things you are looking for (the cname and the address(es) it has resolved to). Using this default serializer will result in only the address being stored.

Here is some mocked up DNS data that illustrates this point (time, cname, addr):

```
2017-01-01 00:00:00Z, host-old.kryo.com, 192.168.1.150
2017-01-02 00:00:00Z, host-new.kryo.com, 192.168.1.150
```

When we store these records via Kryo with this proposed serializer, we end up with:

```
2017-01-01 00:00:00Z, 192.168.1.150
2017-01-02 00:00:00Z, 192.168.1.151
```

Now, when a system goes to pull this data from disk or some other store, we end up with say a `Collection<Record>` where `Record` has two methods: one for the time and one for the address record.

Here is how we would intuitively access this data:

```Java
for (final Record r : storedLookups) {
    final InetAddress addr = r.getInetAddress();
    final Instant time = r.getInstant();

    System.out.println(time.toString() + " " + addr.toString();
}
```

In this case, because the data was restored via the serializer proposed, we end up with the following output:

```
2017-01-01T00:00:00Z host-new.kryo.com/192.168.1.150
2017-01-02T00:00:00Z host-new.kryo.com/192.168.1.150
```

Which means we actually have not only lost data, but what we are looking at is actually wrong. Why? Because we didn't store (nor do we have a good way of doing so without making serialization depend on DNS... and that is problematic too for inter-process communication for processes that have different DNS servers) the host name. Thus, InetAddress did the rDNS lookup and gets the most recent hostname for the address. The historical value is lost.

As such, I strongly feel this should be opt-in with very clear documentation on the caveats of using it.

NathanSweet:
Thanks for explaining the complexity of these serializers. I think they would be best to go into the kryo-serializers project (which may someday make its way into Kryo, but in a separate package and not enabled by default).

PR #516, opened 2017-06-06
Use relatively safe max size of java array

ravwojdyla:
See:
 * https://stackoverflow.com/questions/3038392/do-java-arrays-have-a-maximum-size
 * https://github.com/apache/beam/pull/3189

### DIFF
--- a/src/com/esotericsoftware/kryo/io/ByteBufferInput.java
+++ b/src/com/esotericsoftware/kryo/io/ByteBufferInput.java
@@ -316,7 +316,7 @@ public class ByteBufferInput extends Input {
 	public long skip (long count) throws KryoException {
 		long remaining = count;
 		while (remaining > 0) {
-			int skip = (int)Math.min(Integer.MAX_VALUE, remaining);
+			int skip = (int)Math.min(Util.MAX_SAFE_ARRAY_SIZE, remaining);
 			skip(skip);
 			remaining -= skip;
 		}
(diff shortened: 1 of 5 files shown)


isaki-x:
Is there a reason why the constants in Util aren't marked final? If the code relies on their mutability that is a significant threading issue. I know you are just following the model of what has been done before in that class, but I'm wondering if this should be fixed.

ravwojdyla:
@isaki-x valid question. In the mean time I will fix this in the current PR. Thanks.

src/com/esotericsoftware/kryo/util/Util.java
```diff
@@ -26,6 +26,9 @@
 /** A few utility methods, mostly for private use.
  * @author Nathan Sweet <misc@n4te.com> */
 public class Util {
+	//See: https://stackoverflow.com/questions/3038392/do-java-arrays-have-a-maximum-size
+	static public final int maxSafeArraySize = Integer.MAX_VALUE - 8;
```
magro: As a constant it should probably follow naming conventions for constants?
Alternatively, this could be defined somewhere in the io package and be package private, with the advantage that the public api is not extended (but keeping it here is fine for me as well).

isaki-x: The issue here is that while you are certainly correct, the file being modified does not in fact utilize the standard (so it becomes a style issue: should the file be updated to match the standard or should it be left alone and made at least internally consistent?).

The line after the line the author added is:

`static public boolean isAndroid = "Dalvik".equals(System.getProperty("java.vm.name"));`

This should probably be changed to:

`public static final boolean IS_ANDROID = "Dalvik".equals(System.getProperty("java.vm.name"));`

... or something along those lines. Unless of course you intended for this to be modifiable... then you should probably have set/get and locks/synchronized access or make it an AtomicBoolean (or an AtomicBoolean abstracted by get/set so it can change later).

I do like the idea of minimal visibility and moving this to the IO package is probably the right call regardless of the naming standard and thread safety considerations I have raised.

magro: Oh yeah, you're right, I should have made this final - I don't think that I intended this to be modifiable (I really hope that nobody wants to do this) ;-)

ravwojdyla: @magro thanks for the review. Updated:
 * use constant naming
 * moved to package private in `io`

PTAL

PR #518, opened 2017-06-09
error message for problems with anonymous classes

garfieldnate:
Add suggestions for fixing problems with anonymous classes to the
error that occurs when trying to deserialize them.


magro:
Great, thanks for the PR!

Issue #519, opened 2017-06-09
CompatibleFieldSerializer misses deserializing some fields when new fields have been added to a Class with a large number of fields


barold419:
When a Class exists with a large number of fields, say m (where m > CompatibleFieldSerializer#THRESHOLD_BINARY_SEARCH) is serialized, then n new fields are added to the class before deserialization, then some of the preexisting fields may not be deserialized correctly and the deserialized object will have null values instead of the properly populated values. Further, if this class is part of a larger graph of objects, deserialization may fail completely as the internal deserialization tracking will be broken after missing the previous fields.

magro:
Thanks for reporting! Do you want to submit a pull request?

magro:
Ah, I just spotted #520, great! :-)

magro:
Closed with #520

PR #520, opened 2017-06-09
#519 - Correctly deserialize objects when fields are added to a Class with enough fields to trigger binary search in CompatibleFieldSerializer

barold419:
Correctly deserialize objects when fields are added to a Class with enough fields to trigger binary search in CompatibleFieldSerializer


magro:
Thanks for the PR!

PR #530, opened 2017-08-11
Fix #529: Support serializing the Enum class object

### LINKED ISSUE #529: Serializing the Enum class object
Attempt to serialize the Enum class object triggers a `type cannot be null` error because of the following test `(!type.isEnum() && Enum.class.isAssignableFrom(type))` which is true for the Enum` class object. https://github.com/EsotericSoftware/kryo/blob/64d7784fac8588fcdc4b3574de77b21265c2ad77/src/com/esotericsoftware/kryo/Kryo.java#L493

magro:
When an `Enum.class` instance is serialized (by `ClassSerializer`), this
causes the creation of the registered default serializer (`EnumSerializer`)
with this call chain:
```
ClassSerializer.write
 -> Kryo.writeClass
  -> DefaultClassResolver.writeClass
   -> Kryo.getDefaultSerializer
    -> ReflectionSerializerFactory.makeSerializer(kryo, EnumSerializer, Enum.class)
```

This is done so that `Kryo.writeClass` can write the class name or the id
of the serializer registration.

Until now, the `EnumSerializer` failed instantiation for a type with `null`
enum constants. This however is the case for `Enum.class` itself.
The creation of `EnumSerializer` for `Enum.class` is now allowed, under the
assumption that this happens in the context of the `Enum.class` serialization,
and that `write`/`read` on the created `EnumSerializer` instance will never
be called. This assumption should be safe, since there should never be an instance
of the abstract `Enum` class.

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -490,7 +490,7 @@ public class Kryo {
 			if (Proxy.isProxyClass(type)) {
 				// If a Proxy class, treat it like an InvocationHandler because the concrete class for a proxy is generated.
 				registration = getRegistration(InvocationHandler.class);
-			} else if (!type.isEnum() && Enum.class.isAssignableFrom(type)) {
+			} else if (!type.isEnum() && Enum.class.isAssignableFrom(type) && !Enum.class.equals(type)) {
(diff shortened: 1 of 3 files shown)


pardonut:
Is this fix in the snapshots repo yet?

magro:
@pardonut yes it should.

magro:
@pardonut It got also just released with 4.0.2

PR #531, opened 2017-08-15
Git ignore and repo cleanup

isaki-x:
There were three main changes:

- Ignore IntelliJ project and temp files
- Ignore all VIM tmp files
- Ignore eclipse project files

This also updates to match the current name of .DS_Store on MacOS.

```
$ ls -1 .D*
.DS_Store
```

I also ran a cleanup script to reapply .gitignore to remove any IDE and user specific configuration.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -658,7 +658,7 @@ String value = pool.run(new KryoCallback() {
 
 ## Logging
 
-Kryo makes use of the low overhead, lightweight [MinLog logging library](http://code.google.com/p/minlog/). The logging level can be set by one of the following methods:
+Kryo makes use of the low overhead, lightweight [MinLog logging library](https://github.com/EsotericSoftware/minlog). The logging level can be set by one of the following methods:
 
 ```java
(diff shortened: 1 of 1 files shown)


magro:
Thanks for this! I think @NathanSweet intentionally added eclipse stuff, as one way to propagate code conventions. I'm also not a friend of having IDE specific artifacts in the repo, but then we should have something like checkstyle or some way of auto formatting.

@NathanSweet would you be fine with such an approach and removing eclipse artifacts from the repo?

isaki-x:
If it turns out it isn't ok, Ill roll back the removal part and remove the eclipse portion of the .gitignore changes.

If check style is your goal, you could look at something like setting up travis for this repository so all pull requests are validated against your PMD, Check Style, FindBug, and Code Coverage requirements. At the end of the day is whatever works best for you and your team!

isaki-x:
I rebased my master to deal with merge issues; I have remade this in a branch.

Issue #500, opened 2017-03-12
Jenkins build fails for jdk6


magro:
The latest build on jenkins failed for jdk 6 (running test with previously compiled classes): https://jenkins.inoio.de/job/kryo%20(sun-jdk-6)/263/console

```
java.lang.UnsupportedClassVersionError: hudson/remoting/Launcher : Unsupported major.minor version 51.0
```

This was not caused by a code change but by an update of the jenkins that was done in the meantime. 

The last successful build with jdk 6 https://jenkins.inoio.de/job/kryo%20(sun-jdk-6)/262/console logged an error regarding the java version and it might be the case that the build job fell back to jdk 7, so that tests already have actually been run with jdk 7 (I'm not 100% sure if this was really the case, as some jenkins remoting is involved):

```
[kryo@2] $ /usr/lib/jvm/java-6-oracle/bin/java -cp /var/lib/jenkins/plugins/maven-plugin/WEB-INF/lib/maven3-agent-1.6.jar:/var/lib/jenkins/tools/hudson.tasks.Maven_MavenInstallation/maven3/boot/plexus-classworlds-2.4.jar org.jvnet.hudson.maven3.agent.Maven3Main /var/lib/jenkins/tools/hudson.tasks.Maven_MavenInstallation/maven3 /var/cache/jenkins/war/WEB-INF/lib/remoting-2.53.2.jar /var/lib/jenkins/plugins/maven-plugin/WEB-INF/lib/maven3-interceptor-1.6.jar /var/lib/jenkins/plugins/maven-plugin/WEB-INF/lib/maven3-interceptor-commons-1.6.jar 34662
<===[JENKINS REMOTING CAPACITY]===>channel started
ERROR: [JENKINS-18403] JDK 5 not supported to run Maven; retrying with slave Java and setting compile/test properties to point to /usr/lib/jvm/java-6-oracle
[...]

PR #533, opened 2017-08-21
support java 1.5+

qxo:


isaki-x:
This is arguable a step in the wrong direction. If anything, Kryo should drop support for Java 7. Guava already only builds for 1.8 for example (look at their master pom).

If Kryo really wants to support legacy, unsupported Java versions then I ask that they do what bouncy castle and other frameworks have done and provide a standard build on the oldest supported Java (Java 8) and a legacy build for compatibility with older Java versions. There is no reason to pay the performance and feature cost associated with supporting old, unsupported JVMs for every application that utilizes Kryo when up until this point every JVM using it has been at least Java 7.

Perhaps the right answer is to update the 3.X series to be compatible with 1.5, the 4.X series should remain compatible with 7, and the new 5.X series should support only the oldest supported JVM (Java 8). This will also have the benefit of allowing Kryo to use new language features in its code base.

magro:
I'd agree @isaki-x

Also related: #500

qxo:
ok ,It'll change javaTarget default to 1.7 , leave choice for the library publisher

NathanSweet:
If it doesn't hurt Kryo to support older Java versions, we should do so. I don't believe there is any performance cost with generating bytecodes compatible with older versions of Java. That leaves only the benefit of using newer language features in the Kryo codebase, which is not terribly important for us. If we can easily support newer Java version features, such as how we serialize lamdas, without losing support for older versions of Java, that is good.

FWIW, I personally have Java 7 projects using Kryo which won't be updating to a newer Java any time soon for various reasons.

isaki-x:
This PR is now effectively a no-op. Placing the version in a property isn't a bad idea. However, I tend to avoid using properties without prefixes. It is up to the Kryo owners, but I would probably name the property something like ```kryo.javaTarget```.

PR #537, opened 2017-08-31
README: Change MinLog link away from Google Code

hulbert:
This matches the other link in the README for MinLog which already points to Github


magro:
Thanks!

PR #539, opened 2017-09-08
fix compatible field serializer binary search bugs

lixt:
Fix deserializing bugs using CompatibleFieldSerializer with object which can trigger binary search. Just set the max boundary of binary search to the length of target array. I also update the unit test for this situation.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -98,7 +98,7 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 				// binary search for schemaName
 				int low, mid, high;
 				int compare;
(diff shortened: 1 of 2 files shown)


magro:
Thanks for the PR!

Can you describe in greater detail in which case/scenario an unexpected effect is triggered?
Perhaps a dedicated test for this situation makes sense?

lixt:
This unexpected effect is triggered when binary search is used on object with more than 32 fields.  And the case I have already added into the unit test. Just a little mistake in Kryo's binary search algorithm~

lixt:
Just a glance of the code in CompatibleFieldSerializer.java and you'll know what i'm talking about. With this bug, Kryo losts its backward compatibilty when removing fields in objects with more than 32 fields.

abourov:
This is a fairly major bug, is there a reason why there is no bug report for it, nor has this pull request been added to any release?

PR #545, opened 2017-10-13
remove intermediate byte[] and change visibility to enable override

guoguanglei:
1. remove intermediate byte[] to avoid garbage in ByteBufferOutput
2. change visibility to enable override in Input

### DIFF
--- a/src/com/esotericsoftware/kryo/io/ByteBufferInput.java
+++ b/src/com/esotericsoftware/kryo/io/ByteBufferInput.java
@@ -225,7 +225,8 @@ public class ByteBufferInput extends Input {
 
 	/** @param optional Try to fill the buffer with this many bytes.
 	 * @return the number of bytes remaining, but not more than optional, or -1 if the EOS was reached and the buffer is empty. */
-	private int optional (int optional) throws KryoException {
+	@Override
+	protected int optional (int optional) throws KryoException {
 		int remaining = limit - position;
 		if (remaining >= optional) return optional;
 		optional = Math.min(optional, capacity);
--- a/src/com/esotericsoftware/kryo/io/ByteBufferOutput.java
+++ b/src/com/esotericsoftware/kryo/io/ByteBufferOutput.java
@@ -432,8 +432,9 @@ public class ByteBufferOutput extends Output {
 			if (capacity - position < charCount)
 				writeAscii_slow(value, charCount);
 			else {
-				byte[] tmp = value.getBytes();
-				niobuffer.put(tmp, 0, tmp.length);
+				for (int i = 0; i < value.length(); ++i) {
+					this.niobuffer.put((byte) value.charAt(i));
+				}
 				position += charCount;
 			}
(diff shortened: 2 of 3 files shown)


isaki-x:
I can't comment on the visibility change; I leave that to the repo owners.

However.

This change is a performance decrease; the cost of allocating a pointer on the stack to the backing array of the buffer will be cheaper than invoking methods to access the same data. Using the raw array is the better option. While using the method based access may look nicer, for a library like Kryo where performance is key such convinces are unwarranted.

Calling methods in your for loop declaration is even worse. Take this simple example:

```java
import org.junit.Test;

public class FooTest {
    
    @Test
    public void testLoop() {
        final TestLoopHelper value = new TestLoopHelper(10);
        for (int i = 0; i < value.get(); ++i) {
            System.out.println("Current loop value: " + Integer.toString(i));
        }
    }

    private static final class TestLoopHelper {
        
        private final int x;
        private int counter;
        
        TestLoopHelper(final int x) {
            super();
            this.x = x;
            this.counter = 0;
        }
        
        public int get() {
            System.out.println("Hello world: " + Integer.toString(++this.counter));
            return this.x;
        }
    }  
}
```

The method in loop gets called 11 times for a 10 iteration loop!

```
Hello world: 1
Current loop value: 0
Hello world: 2
Current loop value: 1
Hello world: 3
Current loop value: 2
Hello world: 4
Current loop value: 3
Hello world: 5
Current loop value: 4
Hello world: 6
Current loop value: 5
Hello world: 7
Current loop value: 6
Hello world: 8
Current loop value: 7
Hello world: 9
Current loop value: 8
Hello world: 10
Current loop value: 9
Hello world: 11
```

isaki-x:
@magro I use Kryo for its performance; I'm always watching the PRs and commits to make sure nothing is going to impact my very tight performance constraints.

magro:
@isaki-x awesome! :-)

NathanSweet:
String `getBytes` should generally be avoided. We avoid it in Output so makes sense to do the same in ByteBufferOutput.

ByteBufferInput `optional` is intended to be private API. What reason would there be for overriding it? You are supposed to override `fill` instead.

isaki-x:
@NathanSweet 

This implementation has introduced a bug/compatibility issue. You are casting a char to a byte, which is invalid in Java due to unicode (such as UTF-8) as a character can be more than one byte.

Also, calling charAt() does incur a bounds check on every invocation, so it might be worth benchmarking to see if this route is actually worth avoiding the extra byte[] allocation.

Finally, if you are going to be unicode compatible (which you were, and no longer are, so that's a breaking change) you will end up allocating a new byte[] for every char when you decode the unicode into raw bytes (or have complex code to do the decoding directly instead of using built in classes), so the old way of doing things was probably the best way all things considered.

Thanks!

NathanSweet:
@isaki-x Casting to byte is valid because in this code path every character in the string is guaranteed to be ASCII. I didn't benchmark this change, but String writing was benchmarked for Output `writeString`, which does the same.

isaki-x:
Sweet (no pun intended). Thanks for the info.

PR #546, opened 2017-10-15
Adding import statements to "Quickstart" section.

aidin36:
I add ``import`` statements to the "Quickstart" section, so end user don't have to search around to find where are Kryo classes are.


magro:
Thanks for the PR!

PR #547, opened 2017-10-27
int method readName, Class.forName(className, false, kryo.getClassLoa…

devilyard:
in the method readName, Class.forName(className, false, kryo.getClassLoader) failed with kryo's classLoader, and retry with current one, but Class.forName(className) is tend to load with caller's classLoader, that caller is commonly created by kryo. so i think here use Class.forName(className, false, Thread.currentThread.getContextClassLoader) is seems more suitable.

### DIFF
--- a/src/com/esotericsoftware/kryo/util/DefaultClassResolver.java
+++ b/src/com/esotericsoftware/kryo/util/DefaultClassResolver.java
@@ -36,8 +36,8 @@ public class DefaultClassResolver implements ClassResolver {
 
 	protected Kryo kryo;
 
-	protected final IntMap<Registration> idToRegistration = new IntMap();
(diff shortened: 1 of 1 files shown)


isaki-x:
I tend to agree with your assessment here. However, your generic declarations may not build as Kryo currently uses Java 7 as its build target.

magro:
@devilyard the code you want to change was introduced with #368. Therefore I'd say you should add another classloading attempt with the context classloader.

And as @isaki-x already said, the introduced diamond operators would break the build, please revert these changes.

NathanSweet:
Kryo has a `setClassLoader` method, so users can set the context classloader if they like.
```kryo.setClassLoader(Thread.currentThread.getContextClassLoader());```

PR #560, opened 2017-12-18
Make Map clear resize value configurable

isaki-x:
There was a recent change to use a default size of 2048 when reseting the backing map for resolvers. I have made this a configurable option in the Kryo class. Additionally, I have made several core fields in IndentityObjectIntMap private. This actually causes issues with the build as it detects I have taken something away when it does compatibility checks. If you give me a hint how to fix that, I will. Also, feel free to suggest better names for everything.

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -124,8 +124,6 @@ public class Kryo {
 	static private final int REF = -1;
 	static private final int NO_REF = -2;
 
-	private static final int DEFAULT_IOIM_CLEAR_MAX_SIZE = 2048;
-
 	private SerializerFactory defaultSerializer = new ReflectionSerializerFactory(FieldSerializer.class);
 	private final ArrayList<DefaultSerializerEntry> defaultSerializers = new ArrayList(33);
 	private final int lowPriorityDefaultSerializerCount;
@@ -157,8 +155,6 @@ public class Kryo {
 	private TaggedFieldSerializerConfig taggedFieldSerializerConfig = new TaggedFieldSerializerConfig();
 
 	private StreamFactory streamFactory;
-
-	private int ioimClearSize = DEFAULT_IOIM_CLEAR_MAX_SIZE;
 
 	/** Creates a new Kryo with a {@link DefaultClassResolver} and a {@link MapReferenceResolver}. */
 	public Kryo () {
@@ -1088,15 +1084,6 @@ public class Kryo {
 		this.copyReferences = copyReferences;
 	}
 
-	/**
-	 * This method sets the the size of resolver maps when they are reset/cleared.
-	 *
-	 * @param ioimClearSize
-	 */
-	public void setIdenityObjectIntMapClearSize(int ioimClearSize) {
-		this.ioimClearSize = ioimClearSize;
-	}
-
 	/** The default configuration for {@link FieldSerializer} instances. Already existing serializer instances (e.g. implicitely
 	 * created for already registered classes) are not affected by this configuration. You can override the configuration for a
 	 * single {@link FieldSerializer}. */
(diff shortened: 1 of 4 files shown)


src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -1084,6 +1088,15 @@ public void setCopyReferences (boolean copyReferences) {
 		this.copyReferences = copyReferences;
 	}
 
+	/**
+	 * This method sets the the size of resolver maps when they are reset/cleared.
+	 *
+	 * @param ioimClearSize
+	 */
+	public void setIdenityObjectIntMapClearSize(int ioimClearSize) {
```
NathanSweet: This setting does not belong on the Kryo class, lest it become a junk drawer for various settings.

Why do users need this configurable? When would they want to change it?

src/com/esotericsoftware/kryo/util/IdentityObjectIntMap.java
```diff
@@ -523,6 +523,15 @@ public void ensureCapacity (int additionalCapacity) {
 		if (sizeNeeded >= threshold) resize(ObjectMap.nextPowerOfTwo((int)(sizeNeeded / loadFactor)));
 	}
 
+	/**
+	 * This method returns the current size of the map.
+	 *
+	 * @return The size of the map.
```
NathanSweet: Don't repeat yourself.

src/com/esotericsoftware/kryo/util/IdentityObjectIntMap.java
```diff
@@ -34,11 +34,11 @@
...
 
-	public int size;
+	private int size;
 
-	K[] keyTable;
-	int[] valueTable;
-	int capacity, stashSize;
+	private K[] keyTable;
+	private int[] valueTable;
+	private int capacity, stashSize;
```
NathanSweet: size being private/public should not affect building. The others can be private.

src/com/esotericsoftware/kryo/Kryo.java
isaki-x: I'm against hardcoding values in arbitrary places. As such, I thought a constant would be nice alternative. Then I figured, well, if I'm making it a constant I might as well make it tunable. It is easy enough to go back to just a constant. I'd change the invocation paradigm a bit in this case though. I'll do the change and you'll see what I mean in this PR.

src/com/esotericsoftware/kryo/util/IdentityObjectIntMap.java
isaki-x: But I love hearing the sound of my own voice! I'll fix it.

src/com/esotericsoftware/kryo/util/IdentityObjectIntMap.java
isaki-x: This is the error I see when I mark size private:

```
[INFO] --- clirr-maven-plugin:2.8:check (check-on-compile) @ kryo ---
[INFO] artifact com.esotericsoftware:kryo: checking for updates from sonatype-releases
[INFO] artifact com.esotericsoftware:kryo: checking for updates from sonatype-nexus-snapshots
[INFO] artifact com.esotericsoftware:kryo: checking for updates from central
[INFO] Comparing to version: 4.0.1
[ERROR] 6010: com.esotericsoftware.kryo.util.IdentityObjectIntMap: Accessibility of field size has been weakened from public to private
```

isaki-x:
Changes committed.

NathanSweet:
The size passed to `clear(maxSize)` may not be the same for all situations, so doesn't make sense to have as a constant in the collection class. The `*Resolver` classes are already pluggable, so I think 2048 is a fine default and people can implement their own resolver if changing it becomes necessary.

PR #565, opened 2018-01-02
Support for serialization of very deeply nested object graphs by mean…

tomasz-kuchta:
…s of continuation-based serializers (i.e. by means of continuations passing style)


Erhannis:
What happened to this?  I've actually been using the referenced commit in one of my projects for over a year, but recently made a change that would benefit from some of the newer Kryo generics support.  The recent Kryo builds still crash on deeply nested graphs, which seems like it really ought to be fixed, regardless.

PR #567, opened 2018-01-14
Use public member instead of private field to resolve immutability

MThomassen:
Current com.esotericsoftware.kryo.Serializer::copy function determines immutability through its private field instead of it's public isImmutable() member (which docs actually mention). Since isImmutable() is virtual, this could break the copy api.

Note that this actually occurs in the Scala serializers provided by Twitter's Chill ([Traversable.scala](https://github.com/twitter/chill/blob/master/chill-scala/src/main/scala/com/twitter/chill/Traversable.scala)) and causes lot's of KryoException to be thrown in Apache Flink + Scala usage (one could argue the Twitter serializer implementations are to blame.. or chaning member isImmutable() to be **final** would be a preferred option, but that would break API compatibility)

### DIFF
--- a/src/com/esotericsoftware/kryo/Serializer.java
+++ b/src/com/esotericsoftware/kryo/Serializer.java
@@ -100,7 +100,7 @@ public abstract class Serializer<T> {
 	 * This method should not be called directly, instead this serializer can be passed to {@link Kryo} copy methods that accept a
 	 * serialier. */
 	public T copy (Kryo kryo, T original) {
-		if (immutable) return original;
+		if (isImmutable()) return original;
 		throw new KryoException("Serializer does not support copy: " + getClass().getName());
 	}
 }


isaki-x:
The immutable field is set in the constructor so this isn't really needed; it is arguable that the copy method should have been marked final, but given there are people who have overridden it, this might be the only solution if for some reason you don't want to set it in the constructor.

NathanSweet:
It's typical in Java for methods to not be marked final and anyone overriding a method, especially a getting, should be wary that the method may not be used internally. The javadocs here are on the setter, I'm not sure why anyone would think it a good idea to override the getter. Using only public accessors internally is generally a pain, though this particular scenario it is very simple, so I'm OK with it. In other areas, it's not safe to assume getters will be used internally.

PR #570, opened 2018-02-07
Added ReverseComparator2 serializer, unit tests, and registered as a default serializer

### LINKED ISSUE #569: java.util.Collections$ReverseComparator2 Fails to serialize correctly
```
Class cannot be created (missing no-arg constructor): java.util.Collections$ReverseComparator2
com.esotericsoftware.kryo.KryoException: Class cannot be created (missing no-arg constructor): java.util.Collections$ReverseComparator2
```

There should be a default serializer that handles this wrapper around an existing Comparator correctly.

magneticflux-:
Closes #569


magro:
As written in #569, I think this is not needed but can be solved with `kryo.setInstantiatorStrategy(new Kryo.DefaultInstantiatorStrategy(new StdInstantiatorStrategy()));`.

PR #574, opened 2018-03-21
Add travis build config

magro:
Tests with java8 and java7, on successful build publishes the artifacts to sonatype snapshots.

The inoio jenkins required too much maintenance and became unstable the last days.
We also need travis to be able to build/test with jdk 9/10.
An additional advantage of Travis are PR verification builds, so that it's immediately clear if anything got broken. 

@NathanSweet can you please activate the Travis build according to 2. from the [Getting Started guide](https://docs.travis-ci.com/user/getting-started/#To-get-started-with-Travis-CI)?

### DIFF
--- a/.settings.xml
+++ b/.settings.xml
@@ -1,0 +1,11 @@
+<!-- settings.xml -->
+<settings>
+<servers>
+    <server>
+        <!-- Maven Central Deployment -->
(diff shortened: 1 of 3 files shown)


magro:
@NathanSweet Alternatively you can make me an admin for kryo on github so that I can activate the travis build myself.

Issue #548, opened 2017-10-31
Collections.singleton and friends should not be considered immutable


alexandrebouchard:
By default, Kryo performs shallow copy of instances of Collections.singleton (and singletonList, etc). This is incorrect in cases where the element in the singleton is not immutable, for example a singleton containing an ArrayList.

magro:
@alexandrebouchard Do you want to submit a pull request?

alexandrebouchard:
Sure! Here it is: https://github.com/EsotericSoftware/kryo/pull/575

PR #575, opened 2018-03-21
Fix EsotericSoftware/kryo#548

alexandrebouchard:
Change behaviour of the singleton serializers: contents of the singletons no longer considered immutable.


magro:
Thanks! To prevent regressions it would be awesome to have a test for this.

alexandrebouchard:
My pleasure. Just added test cases.

magro:
Thanks! Can you please squash the commits? Then I will merge.

magro:
Thanks! You could also push -f btw.

magro:
Closed with #576

PR #577, opened 2018-04-01
WIP: Pooled IO objects.

johnou:
180% increase in throughput using pooled Output objects with default buffer size of 4096.

```
Benchmark                       (bufferSize)   Mode  Cnt        Score        Error  Units
KryoIOPoolTest.testOutput                1024  thrpt   40  4175474.459 ± 143755.009  ops/s
KryoIOPoolTest.testOutput                4096  thrpt   40  2371639.643 ±  73842.631  ops/s
KryoIOPoolTest.testPooledOutput          1024  thrpt   40  6643300.248 ± 217470.355  ops/s
KryoIOPoolTest.testPooledOutput          4096  thrpt   40  6653441.255 ± 266002.330  ops/s
```

```
Benchmark                       (bufferSize)   Mode  Cnt        Score       Error  Units
KryoIOPoolTest.testInput                1024  thrpt   40  1087908.503 ± 20258.313  ops/s
KryoIOPoolTest.testInput                4096  thrpt   40   898640.372 ±  7628.268  ops/s
KryoIOPoolTest.testPooledInput          1024  thrpt   40  1194824.270 ± 14490.884  ops/s
KryoIOPoolTest.testPooledInput          4096  thrpt   40  1321546.126 ± 67599.732  ops/s
```

### DIFF
--- a/src/com/esotericsoftware/kryo/pool/KryoIOPoolQueueImpl.java
+++ b/src/com/esotericsoftware/kryo/pool/KryoIOPoolQueueImpl.java
@@ -1,0 +1,57 @@
+/* Copyright (c) 2008, Nathan Sweet
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
+ * conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the distribution.
+ * - Neither the name of Esoteric Software nor the names of its contributors may be used to endorse or promote products derived
+ * from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. */
+
+package com.esotericsoftware.kryo.pool;
+
+import java.util.Queue;
+
+/** A simple {@link Queue} based pool.
+ *
+ * @author Johno Crawford (johno.crawford@gmail.com) */
+abstract class KryoIOPoolQueueImpl<T> {
+
+    private final Queue<T> queue;
+
+    KryoIOPoolQueueImpl(Queue<T> queue) {
+        this.queue = queue;
+    }
+
+    private T borrow(final int bufferSize) {
+        final T element = queue.poll();
+        if (element != null) {
+            return element;
+        }
+        return create(bufferSize);
+    }
+
+    protected abstract T create(final int bufferSize);
+
+    protected void release(final T element) {
+        queue.offer(element);
+    }
+
+    public <R> R run(final KryoIOCallback<T, R> callback, final int bufferSize) {
+        final T element = borrow(bufferSize);
+        try {
+            return callback.apply(element);
+        } finally {
+            release(element);
+        }
+    }
+}
--- a/src/com/esotericsoftware/kryo/pool/KryoInputPool.java
+++ b/src/com/esotericsoftware/kryo/pool/KryoInputPool.java
@@ -1,0 +1,62 @@
+/* Copyright (c) 2008, Nathan Sweet
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
+ * conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the distribution.
+ * - Neither the name of Esoteric Software nor the names of its contributors may be used to endorse or promote products derived
+ * from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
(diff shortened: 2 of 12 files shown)


src/com/esotericsoftware/kryo/pool/KryoIOPoolQueueImpl.java
```diff
@@ -0,0 +1,59 @@
...
+            return element;
+        }
+        return create(bufferSize);
+    }
+
+    protected T create(final int bufferSize) {
+        throw new UnsupportedOperationException("Not implemented");
+    }
+
+    protected void recycle(final T element) {
```
magro: Naming: rename to "release" to be consistent with KryoPool?

src/com/esotericsoftware/kryo/pool/KryoIOPoolQueueImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private T borrow(final int bufferSize) {
+        final T element = queue.poll();
+        if (element != null) {
+            return element;
+        }
+        return create(bufferSize);
+    }
+
+    protected T create(final int bufferSize) {
```
magro: As I understand it, there must be a subclass which provides an implementation. Maybe this class here and this method should be abstract then?

src/com/esotericsoftware/kryo/pool/KryoInputPool.java
```diff
@@ -0,0 +1,67 @@
...
+        private boolean softReferences;
+
+        /** Use {@link SoftReference}s for pooled {@link Kryo} instances, so that instances may be garbage collected when there's
+         * memory demand (by default disabled). */
+        public KryoInputPool.Builder softReferences() {
+            softReferences = true;
+            return this;
+        }
+
+        public KryoInputPool.Builder setMinBufferSize(int minBufferSize) {
```
magro: Change to "minBufferSize" for naming consistency?

src/com/esotericsoftware/kryo/pool/KryoIOPoolQueue.java
```diff
@@ -0,0 +1,25 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. */
+
+package com.esotericsoftware.kryo.pool;
+
+/** @author Johno Crawford (johno.crawford@gmail.com) */
+public interface KryoIOPoolQueue<T> {
+    <R> R run(final KryoIOCallback<T, R> callback, final int bufferSize);
```
magro: Add some documentation?

johnou: What's the Java level? If 8 we could use Function.

src/com/esotericsoftware/kryo/pool/KryoIOPoolQueueImpl.java
johnou: Sgtm

src/com/esotericsoftware/kryo/pool/KryoInputPool.java
johnou: Yep

src/com/esotericsoftware/kryo/pool/KryoIOPoolQueue.java
magro: We're still on java 7

src/com/esotericsoftware/kryo/pool/KryoIOPoolQueueImpl.java
johnou: yep

test/com/esotericsoftware/kryo/pool/KryoInputPoolTest.java
```diff
@@ -0,0 +1,82 @@
...
+                return null;
+            }
+        }, 0);
+        pool.run(new KryoIOCallback<Input, Object>() {
+            public Object apply(Input input) {
+                result[1] = input;
+                return null;
+            }
+        }, 0);
+        assertTrue(result[0] == result[1]);
```
magro: This test should also make the expectation regarding the input state (position etc)  explicit, respectively test if repeated usage actually works as expected.

test/com/esotericsoftware/kryo/pool/KryoOutputPoolTest.java
```diff
@@ -0,0 +1,82 @@
...
+                return null;
+            }
+        }, 0);
+        pool.run(new KryoIOCallback<Output, Object>() {
+            public Object apply(Output output) {
+                result[1] = output;
+                return null;
+            }
+        }, 0);
+        assertTrue(result[0] == result[1]);
```
magro: This test should also make the expectation regarding the output state (position etc)  explicit, respectively test if repeated usage actually works as expected.

johnou:
> There are several Input/Output variants, but the pool only supports Input/Output.

Yes.

> Here and there license headers are missing.

Added.

> Assuming both Kryo and IO objects shall be pooled, should we provide support for this, or is there a use case for just pooling IO?

Any suggestions on how the API should look?

> The README should be extended.

Added some text and examples.

test/com/esotericsoftware/kryo/pool/KryoInputPoolTest.java
johnou: I added some asserts for checking position, buffer will contain the previous payload / allocated byte array and be overwritten next use, not sure if it's worth checking that?

johnou:
```
Benchmark                       (bufferSize)   Mode  Cnt        Score       Error  Units
KryoIOPoolTest.testInput                1024  thrpt  100  1475285.894 ± 29547.202  ops/s
KryoIOPoolTest.testInput                4096  thrpt  100  1485506.370 ± 30023.743  ops/s
KryoIOPoolTest.testPooledInput          1024  thrpt  100  1422686.396 ± 15676.236  ops/s
KryoIOPoolTest.testPooledInput          4096  thrpt  100  1399205.422 ± 23277.763  ops/s
```

not a clear win for Input objects, I will remove the pooling for them.

johnou:
This implementation is actually at risk of buffer overflow, while the previous implementation was not..

magro:
>> Assuming both Kryo and IO objects shall be pooled, should we provide support for this, or is there a use case for just pooling IO?
> 
> Any suggestions on how the API should look?

That would probably not be nice, also the builder of builders would not feel that natural. It's probably better to keep both things separated...

magro:
Regarding the support of multiple `Output` implementations such as `ByteBufferOutput` etc, this would probably mean that for a certain output type a `KryoOutputPool` implementation would be needed (where the implementation is responsible for output creation and for the implementation of `release`). The `Builder` which constructs the output pool instance is then specific for the implementation, right?
To be open for other output pool implementations in the future without having to break binary/source compatibility IMO we'd have to move the builder to the output pool implementation. And the `KryoOutputPool` would have to be changed to s.th. like `KryoOutputPool<O extends Output>`.

WDYT?

NathanSweet:
The benchmarks are essentially measuring allocation vs buffer reuse, with the Kryo serialization just adding noise.

It's a little odd that it's possible to call `borrow(123)` and get back an output or input with less than 123 capacity.

I'm not a fan of the callback interface. I don't see why it's needed -- just obtain, use, and free. I'm also not a fan of having a builder API for such simple classes. Note these two points also apply to KryoPool.

It seems this could all be much simpler with a generic pool that has optional thread safety and optional soft references. This is useful for Kryo, Input, Output, and anything else. No factories: override a method to provide construction and configuration. Put that burden on application code and we no longer have a problem with what type of Input or Output to construct, or the many ways it could be configured. Doing that stuff belong in application code without a pool. IMO there's no reason to make using a pool any different. Wanting to package up creation and configuration into a factory for reuse is separate.

NathanSweet:
See what you guys think:
* Commit 309bcdf3830e49abbb9b6c0f9a29acb564bcf24b
* Docs: https://github.com/EsotericSoftware/kryo#pooling
* Class: https://github.com/EsotericSoftware/kryo/blob/master/src/com/esotericsoftware/kryo/util/Pool.java
We can continue discussion in #602.

Issue #599, opened 2018-06-13
Are there settings of code formatter?


yokotaso:
I want to contribute to kryo project.

I update  Junit3 style test code to Junit4 style.
I want to format java code, but there are no code formatter settings, or code style document.

Are there code formatter settings?

yokotaso:
Thank you for your comment!

PR #600, opened 2018-06-14
Update Junit3 style Test code to Junit4 style Test code

yokotaso:
Update Junit3 Style Test code to Junit4 Sytle Test code.

Althogh Junit5 is already GA, I update to Junit4.
Because kryo supports java7, and Junit5 requires java8~.

`@Rule` `@RunWith` is removed at Junit5, so I tried to write test without `@Rule` `@RunWith`

So, It is needed to update Junit5 when kryo stop supporting java7,
But Updating Junit5 from Junit4 is easier than Updating Junit5 from Junit3, I think.

I formatted [this settings](https://github.com/EsotericSoftware/kryo/issues/599), But this settings has not settings about import order and 
unused import rule. if import order is wrong, feel free to comment about import order.
(unused import will be removed. it is right, isn't it?)

If you have free time, Cloud you review it?
Finally, Thank you for developing wonderful library!


yokotaso:
> I'd prefer to keep things as concise as possible. Also it's usually best to change as few things as possible, especially for large commits.

Sorry.  I am totally agree with you.

> Why not continue to extend KryoTestCase (which doesn't need to extend TestCase) so support

Though I saw some test codes, I found code which overwrite parent class protected field value.
KryoTestCase will have state in test method. 
I think that it is not clear, misleading and difficult to read.

I want to remove state in test method as possible as I can.

> I'd prefer to move to @before, @test, @after annotations, expected exceptions, static assert import, parameterization, etc without unnecessary changes, or at least only with changes that make the tests smaller, not larger.

Sorry. I should make small change set which move to use `@Before` `@Test` `@After`

> Unfortunately the timing for a commit which touches so much existing code isn't great

Oh sorry, I did not notice.
Though It is ok this PR will be close, May I re-send PR which move to use `@Before` `@Test` `@After`?

I use kryo too heavily in my project. As token of thanks, I want to help you something.

NathanSweet:
kryo-5.0.0-dev has been merged! Please feel free to submit a new PR. It would be best to start with only changing to before, test, after annotations. Cheers!

NathanSweet:
I almost forget, regarding the import order, it seems it is not part of the formatter but it should be in the per project settings of the included Eclipse project files. You may have to use Eclipse for that, or you can submit a PR and it only takes a few seconds for me to run the formatter if that is easier.

Issue #397, opened 2016-03-01
Update tests to use junit 4 (instead of 3)


magro:
Tests inheriting from KryoTestCase are run by junit 3 runner which lacks some features of junit 4 (e.g. does not support Assume to skip tests, which was introduced in 4.4).

Changing KryoTestCase to run with junit 4 would need changes to various tests: assert\* would have to be imported,  @Test would have to be added and @Before would have to be used.

NathanSweet:
#600

PR #601, opened 2018-06-15
Move to Junit4 Style test code (Part 1).

yokotaso:
- test code in package com.esoterisoftware.kryo and com.esoterisoftware.kryo.util
- Add @Test, @Before annotations.
- Use assertion method in org.junit.Assert as possible as I can.
- Change to use Assert.assertArrayEquals for byte array comparison test.

### DIFF
--- a/test/com/esotericsoftware/kryo/CopyTest.java
+++ b/test/com/esotericsoftware/kryo/CopyTest.java
@@ -21,12 +21,17 @@ package com.esotericsoftware.kryo;
 
 import java.util.ArrayList;
 
+import org.junit.Before;
+import org.junit.Test;
+
 public class CopyTest extends KryoTestCase {
+	@Before
 	public void setUp () throws Exception {
 		super.setUp();
 		kryo.setRegistrationRequired(false);
 	}
 
+	@Test
 	public void testBasic () {
 		ArrayList test = new ArrayList();
 		test.add("one");
@@ -38,6 +43,7 @@ public class CopyTest extends KryoTestCase {
 		assertEquals(test, copy);
 	}
 
+	@Test
 	public void testNested () {
 		ArrayList test = new ArrayList();
 		test.add("one");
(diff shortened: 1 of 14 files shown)


NathanSweet:
closes #397

NathanSweet:
Looks good, thanks!

I added an Assert static import where possible, just now.

Would you like to discuss changes for part 2? It's better to discuss first than to do the work and not have it match the Grand Vision. :) I guess next is adding annotations to the rest of the classes, then making KryoTestCase not extends TestCase. We should make sure the same number of tests are still executed after that change.

> Though I saw some test codes, I found code which overwrite parent class protected field value.
> KryoTestCase will have state in test method.
> I think that it is not clear, misleading and difficult to read.

It's true KryoTestCase has state, but it's very convenient. Almost all tests need that state, and it's otherwise safe to ignore. In general I don't mind cutting some corners to make tests less code and easier to write and maintain, which I wouldn't do with the main library. I'm more concerned about those than with readability for the tests.

yokotaso:
Thank you for review and marge PR.

> Would you like to discuss changes for part 2?

Yes.  That's as you say, I planned :)
I planned add juni4 style annotations and making KryoTestCase not extends TestCase.

Additionaly, I want to change deprected method at junit4

- [assertEquals as to double] (https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html#assertEquals(double,%20double))

Those method are deprecated and recommended use [this assertion method](https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html#assertEquals(double,%20double,%20double))

So I planned to add `KryoAssert.assertDoubleEquals` and modify metod which is deprecated.

> It's true KryoTestCase has state, but it's very convenient. Almost all tests need that state, and it's otherwise safe to ignore. In general I don't mind cutting some corners to make tests less code and easier to write and maintain, which I wouldn't do with the main library. I'm more concerned about those than with readability for the tests.

OK.It's all right. I understand your claim :)

PR #603, opened 2018-06-16
Move to Junit4 Style test code (Part 2).

yokotaso:
- Add Junit4 style annotation @Before,@Test.
- Change public method to private if it is possible.
- Stop extending TestCase class, If it is possible.
- Because There are deprecated methods in junit4,
  I add static assertion class which is simple wrapper class.

I will Making KryoTestCase stop extending TestCase class at Part 3.

### DIFF
--- a/test/com/esotericsoftware/kryo/KryoAssert.java
+++ b/test/com/esotericsoftware/kryo/KryoAssert.java
@@ -1,0 +1,21 @@
+package com.esotericsoftware.kryo;
+
+import static org.junit.Assert.*;
+
+/**
+ * Junit Assert wrapper methods class
+ */
+public class KryoAssert {
+
+    private KryoAssert() {
+
+    }
+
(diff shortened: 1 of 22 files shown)


NathanSweet:
Thanks!

> I will Making KryoTestCase stop extending TestCase class at Part 3.

I'm not sure I'll like this change. KryoTestCase is very convenient.

There's a few tests that expect an exception that could be improved.

yokotaso:
Thank you for merge PR.

> I'm not sure I'll like this change. KryoTestCase is very convenient.

Yes. I was convinced that KryoTestCase is very convenient, as you think.
But KryoTestCase extends TestCase. So I want to remove:

```
class KryoTestCase extends TestCase
                  ^^^^^^^^^^^^^^^^^
                   I want to remove it.
`````

> There's a few tests that expect an exception that could be improved.

OK. I try to improve suitable assertion with next PR.

PR #605, opened 2018-06-17
 Move to Junit4 Style test code(Part3)

yokotaso:
Move to Junit4 Style test code(Part3). It is Final.

- Stop extending `org.junit.TestCase` at `KryoTestCase`. And I don't remove `KryoTestCase` :)
- Use assertion methods in `org.junit.Assert`, insted of methods in `org.junit.TestCase`
- Use suitable assertion method like `assertNull` `assertNotSame` `assertSame` `assertTrue` `assertFalse` insted of `assertEquals` if possible
- Remove unused throws statement at method signature.
- Use `@Test(expected = ExceptionClass.class)` where test expects to thrown exception
  as long as test code will be simple. 
Because Junit4 hasn't suitable assertion method for testing exception thrown.
And At Junit5, good assertion method like `assertThrown` is introduced.
So, I think that it is not reasonable that I modify test code now where I can write more clearly with using Junit5 assertion methods. 
Therefore I modify test code which expect to throw exception as long as I can modify simple.

### DIFF
--- a/test/com/esotericsoftware/kryo/CopyTest.java
+++ b/test/com/esotericsoftware/kryo/CopyTest.java
@@ -19,6 +19,8 @@
 
 package com.esotericsoftware.kryo;
 
+import static org.junit.Assert.*;
+
 import java.util.ArrayList;
 
 import org.junit.Before;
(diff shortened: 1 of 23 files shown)


NathanSweet:
Thanks!

Issue #626, opened 2018-10-08
Support for Java 11


nirupamarachuri:
Is there any release planned for kryo-serializers and kryo-shaded that supports Java 11? If yes, when is it planned to release?

NathanSweet:
@nirupamarachuri Kryo works fine with Java 11. When reflection is disallowed (note it is allowed in Java 11 by default), many Kryo serializers will no longer work. Kryo itself should still work fine, but you would need to use serializers that don't rely on reflection. I don't personally have a need to use Java with reflection disabled, so I don't currently have plans to work on alternate serializers.

@free2bcreative It's just a warning. You can read more about it here:
https://groups.google.com/forum/#!topic/kryo-users/bcsag2KfbvA

lucianoRM:
I am also interested in Java 11 support. 
I see the following error when trying to serialize a java.io.FileInputStream:

```java.lang.IllegalArgumentException: Unable to create serializer "com.esotericsoftware.kryo.serializers.FieldSerializer" for class: java.io.FileCleanable
Serialization trace:
cleanup (java.io.FileDescriptor)
fd (java.io.FileInputStream)
	at org.mule.tooling.client.tests.integration.tooling.client.KryoTestCase.serializeBufferedInputStream(KryoTestCase.java:32)
Caused by: java.lang.IllegalArgumentException: Unable to create serializer "com.esotericsoftware.kryo.serializers.FieldSerializer" for class: java.io.FileCleanable
	at org.mule.tooling.client.tests.integration.tooling.client.KryoTestCase.serializeBufferedInputStream(KryoTestCase.java:32)
Caused by: java.lang.reflect.InvocationTargetException
	at org.mule.tooling.client.tests.integration.tooling.client.KryoTestCase.serializeBufferedInputStream(KryoTestCase.java:32)
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field jdk.internal.ref.PhantomCleanable jdk.internal.ref.PhantomCleanable.prev accessible: module java.base does not "opens jdk.internal.ref" to unnamed module @37f1104d
	at org.mule.tooling.client.tests.integration.tooling.client.KryoTestCase.serializeBufferedInputStream(KryoTestCase.java:32)
```

I debugged the code and the error is indeed because "jdk.internal.ref" is not opened in java.base module. 
[...]

PR #637, opened 2018-12-28
Add automatic module name to jar manifest

magro:
For the motivation see https://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html

reflectasm and minlog are updated to versions that provide an
Automatic-Module-Name as well.

objenesis is not updated, because it only provides Automatic-Module-Name
with 3.0.1, which requires java 1.8 (since 3.0.0) and therefore would
require kryo to switch from 1.7 to 1.8.

Refs #626

### DIFF
--- a/pom-main.xml
+++ b/pom-main.xml
@@ -18,7 +18,7 @@
 		<dependency>
 			<groupId>com.esotericsoftware</groupId>
 			<artifactId>reflectasm</artifactId>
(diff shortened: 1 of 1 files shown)


magro:
@NathanSweet Or should we update objenesis as well and move from 1.7 to 1.8 (as the blog post recommends to not depend on a lib without an explicit module name)?

Issue #650, opened 2019-02-22
Major breaking changes while keeping the same package


masc3d:
reading migration guide.

> the best approach is to load the old data somehow, then write it with the newer version. For example, by juggling classloaders the data could be loaded with v2, then written with v5.

I fail to see how this is the best or even a good approach. Could you please elaborate?

As far as I see it's best practice for java libs in general to simply move to a new package for major / breaking updates like this eg. `com.esotericsoftware.kryo.v5`, so everybody is happy with smooth migrations.

NathanSweet:
I don't believe it is common practice with Java libraries to rename packages for every binary incompatible version of a library. I do agree a version in the package could make it easier to migrate data. Dropping in a different major version Kryo JAR is never going to "just work" with old data, though that isn't the only use case. It may be that you don't have persistent data (eg, only short lived network communication) or you are starting over without existing data.

Does anyone have thoughts on renaming Kryo v5 classes to a `v5` package? @magro @romix @serverperformance ?

masc3d:
> I don't believe it is common practice with Java libraries to rename packages for every binary incompatible version of a library

not every, but when it makes sense.
and it does on major breaking api changes and even more so on breaking formats.

many of modern java libs do it eg [rxjava](https://github.com/ReactiveX/RxJava), [retrofit](https://github.com/square/retrofit), [swagger](https://github.com/swagger-api/swagger-core) just to name a few.

in a distributed environment where kryo is protocol, supporting v4 and v5 side-by-side is at least temporarily required for a smooth transition.

magro:
As I see it a common motivation for different package names is to prevent a diamond dependency problem. Assuming an application uses different higher level libraries that are using / working with different versions of an underlying lib, updating one of the higher level libs can cause compatibility issues with the underlying lib. Let's take netty as an example: if an application uses a cassandra driver (which uses netty) and async-http-client (using netty as well) it helps if netty changes its package name _and_ group/artifact id on major updates, because the application then can update the cassandra driver or async-http-client independently without running into conflicts with netty (because the high level libs in fact do no longer depend on two different libraries). Another solution for this is shading, i.e. the cassandra driver or async-http-client could come with a shaded netty. In case of kryo we had this issue with asm, why we provided a kryo artifact with a shaded asm. 
In my view that diamond dependency problem is the most common motivation for changing the package name and artifact id on breaking changes. 

For higher level libs only and directly used by application code in general I don't see much value for changing the package and artifact id. 

[...]

PR #652, opened 2019-02-24
Add versioned kryo artifact (currently kryo5)

magro:
This adds a completely self-contained (dependency free) kryo artifact that can be
used together with another major kryo version at the same time.

Therefore all dependencies are shaded to prevent potential conflicts.
The asm (shaded by reflectasm) that was placed into `com.esotericsoftware.asm`
is moved to `com.esotericsoftware.kryo.kryo5.asm` in order to prevent
conflicts.
While the artifact is produced by depending on kryo (and relocating the kryo package), the produced/published artifact/pom does _not_ depend on kryo, so that it's a completely different artifact (apart from sharing the group id).

The final artifact id is `kryo5`, similarly the Automatic-Module-Name (JPMS).

An OSGi bundle is not published for the versioned artifact, because OSGi handles the things that motivate the versioned jar, and OSGi users are therefore expected to use the OSGi bundle of the "unversioned"/default artifact. Apart from this, producing a valid OSGi bundle for the versioned artifact was not possible out of the box.

The README should still be extended/adjusted to mention this artifact.

Refs #650


develar:
Will be great if issue will be closed, because otherwise it forces to publish own version of kryo.

NathanSweet:
Sorry I'm so late to respond!

PR #658, opened 2019-02-28
Added global registration

Shaunwild97:
Added two new methods

- registerGlobal(Class)
- unregisterGlobal(Class)

Which both save a list of registered classes to a static private member. When a new `Kryo` instance is created, these classes will be registered automatically.


isaki:
Some quick thoughts before the Kryo team takes a peek:

- This is not thread safe; modifiable globals should be thread-safe.
- You may have an issue as this will not backfill existing Kryo instances; this may be counterintuitive given that when you register currently, it affects the existing instance. Not a show stopper by any means but certainly something to think about.

Shaunwild97:
Yea, the thread safety part is something I didn't think about. Good call, I'll sort that. 

As for the back filling, I think that is just a quirk of doing it on the class, instead of the instance. The idea is that this code will run long before the Kryo instances are ever created.

Shaunwild97:
Not that it matters now, but Kryo is not thread safe anyway. See: https://github.com/EsotericSoftware/kryo#thread-safety

NathanSweet:
Sorry, I am not a fan of this addition. Globals are generally a bad idea, but the main reason is that it adds extra noise to the Kryo API for something that is easily done in application code. It is unusual to create many Kryo instances, typically one per thread, and even then the creation is almost always centralized, eg:
https://github.com/EsotericSoftware/kryo#thread-safety

PR #665, opened 2019-04-01
Fix: Call reset before offering

julien-may:
Ensure to reset the object before offering it to the pool to not obtain a not freed object in case of multi threading


magro:
Thanks for the PR!

PR #667, opened 2019-04-11
remove Systen.out.println since it pullutes the console logs

vivekmalhotragithub:


magro:
Thanks for the PR, I just released kryo-5.0.0-RC4 with this fix.

Issue #663, opened 2019-03-22
jvm crash when field type changed


HiwayChe:
JavaBean:
>
    public class Person implements Serializable {
    private static final long serialVersionUID = 1L;
    private String id;
    private String name;
    //getter, setter
    }
>

SerializeUtil:
>     
      static Pool<Kryo> kryoPool = new Pool<Kryo>(true, false, 100) {
        @Override
        protected Kryo create() {
            Kryo kryo = new Kryo();
            UnmodifiableCollectionsSerializer.registerSerializers(kryo);
            kryo.setInstantiatorStrategy(new DefaultInstantiatorStrategy(new StdInstantiatorStrategy()));
            kryo.setReferences(false);
            kryo.setRegistrationRequired(false);
            kryo.setDefaultSerializer(CompatibleFieldSerializer.class);
            return kryo;
        }
    };

serialize a `person` object to file, note that `id` field is now `String`  type, after then change `id` to `Long` type, deserialize from file to a `Person` object, it succeed, and you can see a `String` value `hacker` was assigned to a `Long` field in the below picture(Line 74),
![image](https://user-images.githubusercontent.com/5837440/54816930-93e9ea80-4cd0-11e9-9e55-ac13a7959b06.png)

when code execute to line 77, jvm crashed, maybe I should not change `id` type from `String` to `Long`, but if kryo throws exception instead of crashing jvm will be more elegant.

Env:jdk8, kryo 5.0.0-RC1, kryo-serializers 0.45(I think this issue has nothing to do with kryo-serializers)

isaki:
I've never seen the JVM crash for improper payloads; normally I get a KryoException (which is an unchecked exception) under Kryo 4.0.2. Do you have a stack trace for the crash?

HiwayChe:
> I've never seen the JVM crash for improper payloads; normally I get a KryoException (which is an unchecked exception) under Kryo 4.0.2. Do you have a stack trace for the crash?

I'll try Kryo 4.0.2 and watch the result, also provide reproduction code and jvm crash report later this afternoon.

HiwayChe:
> I've never seen the JVM crash for improper payloads; normally I get a KryoException (which is an unchecked exception) under Kryo 4.0.2. Do you have a stack trace for the crash?

https://github.com/HiwayChe/kryotest
I have uploaded reproduction code in github, you can run Main.testKryo5 to reproduce the issue,  and I write reproduction instruction in code comment.
[...]

PR #670, opened 2019-04-17
Fix JVM Crash as reported in #663

isaki:
This addresses two issues in the CompatibleFieldSerializer.

1) The reported issue; the code now validates that the type read from
the bytecode is compatible with the type read from the serialized data
and skips it/throws (depending on chunk mode) if the type is not
compatible.

2) In implementing the above check, the unit tests began failing. It
turns out that the write method in CompatibleFieldSerializer had a field
cache corruption for primitive types due to autoboxing. This has been
mitigated by checking if the cached type is a primitive before
overwriting it with a value type. Given the primitive types have
dedicated CachedField implementations, this is perfectly safe and has no
binary compatibility concerns.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -88,6 +88,9 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 				Class valueClass = null;
 				try {
 					if (object != null) {
+						// This breaks primitives; it boxes them. As such we need to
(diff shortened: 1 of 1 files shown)


magro:
@isaki Many thanks for your PR! Because of your last comment in #663: is this ready to be merged?

isaki:
This commit is ready to be merged and all tests pass.

isaki:
Not trying to be annoying, but have you had a chance to review this change? Thanks!

NathanSweet:
Sorry for the delay! Much thanks for digging into the fix. Please see #663 for the resolution.

PR #672, opened 2019-05-03
Avoid NPE writing StringBuilder/StringBuffer

candrews:
If a null StringBuilder or StringBuffer is provided, a NullPointerException occurs. Since these serializers both do `setAcceptsNull(true)`, a null is passed to their `write` methods which needs to be handled.


NathanSweet:
Good catch, thanks!

Issue #654, opened 2019-02-26
ArrayIndexOutOfBoundsException when serializing generic classes


theigl:
Using Kryo 5.0.0-RC3-SNAPSHOT, I get the following Stacktrace when running the test:

```
com.esotericsoftware.kryo.KryoException: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
Serialization trace:
input (SerializeTest$StringSupplierContainer)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:92)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.write(FieldSerializer.java:107)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:630)
```

```
import com.esotericsoftware.kryo.Kryo;
import com.esotericsoftware.kryo.io.Output;
import org.junit.jupiter.api.Test;

import java.io.Serializable;
import java.util.function.Supplier;

import static org.assertj.core.api.Assertions.assertThat;

public class SerializeTest {

	@Test
	void serialize() {
		final Kryo kryo = new Kryo();
		kryo.setRegistrationRequired(false);

		final Output output = new Output(1024);
		kryo.writeClassAndObject(output, new StringSupplierContainer());
		final byte[] result = output.toBytes();

		assertThat(result).isNotNull();
	}

	static class EmptyStringSupplier implements Supplier<String>, Serializable {

		@Override
		public String get() {
			return "";
		}
	}

	static class StringSupplierContainer extends SupplierContainer<String> {

		StringSupplierContainer() {
			super(new EmptyStringSupplier());
		}
	}

	static class SupplierContainer<T> {

		private final Supplier<T> input;

		SupplierContainer(Supplier<T> input) {
			this.input = input;
		}
	}
}
```

Possibly related to #648

c14s:
With RC4, I'm also/still getting this ArrayIndexOutOfBoundsException.

Looks that in Generics.java, the method pushTypeVariables() should just return 0, if the hierarchy is empty.

With this change, the test mentioned above is passed successfully.
Or would this change just hide the root cause?

```
public int pushTypeVariables (GenericsHierarchy hierarchy, GenericType[] args) {
           
    if (hierarchy.total == 0) {
        return 0; // no type parameters, e.g. for String
    }
 [...]
```

egorlitvinenko:
And
`
java.lang.NullPointerException: null
[...]

Issue #655, opened 2019-02-26
Unable to resolve type variable


theigl:
Using Kryo 5.0.0-RC3-SNAPSHOT, I get the following Stacktrace when running the test:

```
om.esotericsoftware.kryo.KryoException: Unable to resolve type variable: T

	at com.esotericsoftware.kryo.util.GenericsUtil.resolveTypeVariable(GenericsUtil.java:112)
	at com.esotericsoftware.kryo.util.GenericsUtil.resolveType(GenericsUtil.java:41)
	at com.esotericsoftware.kryo.util.Generics$GenericType.initialize(Generics.java:294)
	at com.esotericsoftware.kryo.util.Generics$GenericType.<init>(Generics.java:259)
	at com.esotericsoftware.kryo.serializers.CachedFields.addField(CachedFields.java:139)
	at com.esotericsoftware.kryo.serializers.CachedFields.rebuild(CachedFields.java:99)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.<init>(FieldSerializer.java:82)
	at com.esotericsoftware.kryo.SerializerFactory$FieldSerializerFactory.newSerializer(SerializerFactory.java:124)
	at com.esotericsoftware.kryo.SerializerFactory$FieldSerializerFactory.newSerializer(SerializerFactory.java:108)
	at com.esotericsoftware.kryo.Kryo.newDefaultSerializer(Kryo.java:395)
	at com.esotericsoftware.kryo.Kryo.getDefaultSerializer(Kryo.java:380)
	at com.esotericsoftware.kryo.util.DefaultClassResolver.registerImplicit(DefaultClassResolver.java:87)
	at com.esotericsoftware.kryo.Kryo.getRegistration(Kryo.java:500)
	at com.esotericsoftware.kryo.util.DefaultClassResolver.writeClass(DefaultClassResolver.java:110)
	at com.esotericsoftware.kryo.Kryo.writeClass(Kryo.java:532)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:627)
```

```
import com.esotericsoftware.kryo.Kryo;
import com.esotericsoftware.kryo.io.Output;
import org.junit.jupiter.api.Test;

import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;
import java.util.function.Supplier;

import static org.assertj.core.api.Assertions.assertThat;

public class SerializeTest {

	@Test
	void serializeClassList() {
[...]

PR #683, opened 2019-07-30
 Attempt to work around problems with generics

MarcMil:
This PR allows to set a different generic saving strategy (i.e. none). I cannot see a difference in the result of the test cases (apart from the expected lengths not matching).
The new behavior is entirely optional and fixes https://github.com/EsotericSoftware/kryo/issues/654 as per test case.

This is like a revival of
https://github.com/EsotericSoftware/kryo/pull/433
for which support seems to have been dropped.

While this PR is working, I do appreciate comments and ideas from maintainers who know this code better than me and might have an idea where this might break.


MarcMil:
And yet again, Travis CI has a problem rather with the environment than with this PR:
```
Installing oraclejdk8
$ export JAVA_HOME=~/oraclejdk8
$ export PATH="$JAVA_HOME/bin:$PATH"
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
Ignoring license option: BCL -- using GPLv2+CE by default
install-jdk.sh 2019-07-17
Expected feature release number in range of 9 to 14, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```

MarcMil:
I contacted the people from travis-ci.
Here's their response:
```
Thank you for getting in touch and our apologies for the hassle here.
As Ubuntu 14.04 reached EOL on April 30th, 2019, we've been slowly rolling out changes to make Xenial builds the default instead of Trusty. How does this impact you? If you haven't explicitly set dist: trusty in your .travis.yml file, your builds may be routed to Xenial instead of Trusty. Sorry for the inconvenience.
To get a better understanding of the reasoning behind this migration, we invite you to look at our blog post: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment.
As previously mentioned, if your builds are suddenly failing due to this change, you can explicitly set dist: trusty to route your builds back to the Trusty distribution.
Hopefully, this helps and don't hesitate to follow up if you have any questions or concerns we can address. We'd be happy to help!
```

maxisvest:
Hello @MarcMil , thanks for this P, I have a question, why you delete the `optimizeGenerics` switch?

dano:
FWIW, here's the commit that dropped support for disabling generics optimizations: https://github.com/EsotericSoftware/kryo/commit/167ef87457ea8fa624e87312ca7680ff97a1bb92
It doesn't really shed any light on why it was dropped, though.

theigl:
@MarcMil: Could you please merge the current master into your branch? I added 3 test-cases for currently known generics issues and it would be great if you could verify that they also work with your code.

test/com/esotericsoftware/kryo/KryoTestCase.java
```diff
@@ -191,8 +191,6 @@ public Input createInput (byte[] buffer) {
 		object2 = kryo.readClassAndObject(input);
 		doAssertEquals(object1, object2);
 		if (checkLength) {
-			assertEquals("Incorrect number of bytes read.", length, input.total());
```
theigl, line 194: Why did you remove these assertions?

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
```diff
@@ -129,6 +135,47 @@ public void testNotPassingToSuper () {
 		kryo.register(DontPassToSuper.class);
 		kryo.copy(new DontPassToSuper());
 	}
+	
+	
+
+	@Test
+	public void testComplicatedGenerics() {
```
theigl, line 142: I have replaced this test-case with a more concise version in #717. Please merge the current master into your branch and verify that all three tests from #717 pass.

theigl:
Please also add an entry to the readme for the new configuration option and briefly explain in what cases it might make sense to opt out of generics optimization and what potential effect it has on binary size, performance and robustness.

In my understanding it should have similar effects as [described in the older PR](https://github.com/EsotericSoftware/kryo/pull/433#issuecomment-225605018), i.e. increased performance and robustness, but larger binary size. Is that correct?

test/com/esotericsoftware/kryo/KryoTestCase.java
MarcMil: Good question, I don't remember anymore, so I reverted that file and the tests still work.

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
MarcMil: I've pulled from the master branch and reverted my changes on that file. The tests pass.

MarcMil:
> Please also add an entry to the readme for the new configuration option and briefly explain in what cases it might make sense to opt out of generics optimization and what potential effect it has on binary size, performance and robustness.
> 
> In my understanding it should have similar effects as [described in the older PR](https://github.com/EsotericSoftware/kryo/pull/433#issuecomment-225605018), i.e. increased performance and robustness, but larger binary size. Is that correct?

You are correct. Do you have a suggestion on where exactly to put this information in the README?
There is the section on generics, however, it is meant for implementers of serializers. But even if the way generics are handled can be changed now, it is probably nonetheless a bad idea to encourage people to ignore the semantics of a generic handler just because they use the no-op generic handler.
Since this is a configuration to configure a trade-off between speed/reliability and size, I feel like it should belong into a more general section. What do you think?

theigl:
> Since this is a configuration to configure a trade-off between speed/reliability and size, I feel like it should belong into a more general section. What do you think?

Good point. It probably fits best as a section under "Reading and writing objects".  @magro do you have an opinion on this?

For now, it is might be enough if you just add JavaDoc to `Kryo.setGenerics()` that explains that setting `NoGenericHandler` can be used to disable generics handling and describe the speed/size tradeoff as JavaDoc on `NoGenericHandler`.

MarcMil:
> For now, it is might be enough if you just add JavaDoc to `Kryo.setGenerics()` that explains that setting `NoGenericHandler` can be used to disable generics handling and describe the speed/size tradeoff as JavaDoc on `NoGenericHandler`.
Done.

magro:
IMHO mentioning this in a separate generics related section in "Reading and writing objects" would make sense: while some users may check what the Kryo class provides, other users might just look in the Readme for help on their issue.

theigl:
Please also add at least one roundtrip test that uses the `NoGenericHandler` so we can be sure that this doesn't get broken accidentally by future changes.

theigl:
I don't really like the `Handler` part of class names, but I'm not sure I can come up with something better.

Maybe something like this?

```
GenericsStrategy
- Generics
- NoGenerics
```

I don't want to start a naming discussion though, so please feel free to ignore this if you are happy with the current class names.

MarcMil:
> Please also add at least one roundtrip test that uses the `NoGenericHandler` so we can be sure that this doesn't get broken accidentally by future changes.

Done; I've adapted the test cases so that *all* roundtrip tests use both implementations for testing.

> GenericsStrategy
I would have named it strategy as well, since it essential an instance of the strategy design pattern. To keep compatibility with the old implementation regarding class names, I've left the name as Handler, because that was the previously chosen name.
I've changed the names now; I also think the name strategy is more appropriate for this case.

test/com/esotericsoftware/kryo/KryoTestCase.java
```diff
@@ -173,6 +175,12 @@ public Input createInput (byte[] buffer) {
 
 	/** @param length Pass Integer.MIN_VALUE to disable checking the length. */
 	public <T> T roundTripWithBufferFactory (int length, T object1, BufferFactory sf) {
+		roundTripWithBufferFactory(length, object1, sf, NoGenericsStrategy.INSTANCE);
+		return roundTripWithBufferFactory(length, object1, sf, new DefaultGenericsStrategy(kryo));
+	}
+
+	/** @param length Pass Integer.MIN_VALUE to disable checking the length. */
+	public <T> T roundTripWithBufferFactory (int length, T object1, BufferFactory sf, GenericsStrategy genericStrategy) {
```
theigl, line 183: It seems you never actually apply the `GenericsStrategy` to the Kryo instance. If you did, I guess that most of the tests would fail because the `length` does not match when `NoGenericsStrategy` is used.

While I like the idea of executing as many tests as possible with the new strategy, it would require two different `length` parameters for every call and is not very intuitive. I think it would be easier for now if you just craft a single test-case that calls all or most of the methods on `NoGenericsStrategy` and put it in `GenericsTest`.

@magro: WDYT?

MarcMil: Oh, you're right, my fault. Hmm... So, what we could do is using both strategies for all non-generic test cases. The generic test cases can specify the strategy to use.
I think, currently there is no test case faiing with the current default generics strategy, right? So the previous test cases could specify this strategy directly.

theigl: > So, what we could do is using both strategies for all non-generic test cases. The generic test cases can specify the strategy to use.

It would probably be enough to create an additional `roundTrip` method with two length parameters and use that explicitly for all generics tests. I looked through the previous implementation of `setOptimizedGenerics` and I think it was only used for generics tests as well. (https://github.com/EsotericSoftware/kryo/commit/167ef87457ea8fa624e87312ca7680ff97a1bb92)

> I think, currently there is no test case faiing with the current default generics strategy, right? 

Correct. Some additional tests were merged into master yesterday, so you might want to pull the changes into your branch.

theigl: But let's wait for feedback from @magro as well.

theigl:
I gave this issue some more thought while out for a walk.

What we want to achieve with this PR is to restore the ability to disable generics optimization that has been enabled by default in Kryo 5. This is important because there are cases where this optimization fails (#654, #655, #721 etc). It also might make sense to disable it to trade increased serialization performance for larger binary size.

This PR achieves that in quite an elegant way, but it exposes a lot of internal API (the whole of `Generics`) that will make it harder to evolve this API in the future. 

The question is this: Will users actually want to implement their own `GenericsStrategy`? Or is it enough to have two default implementations. One that optimizes generics, and one that doesn't.

As far is we know, there are no open issues that require custom implementations of `GenericsStrategy`.

So I would suggest that we keep the strategies internal to Kryo and drop `Kryo.setGenerics()` in favor of `Kryo.setOptimizeGenerics(boolean optimizeGenerics)` that internally sets the appropriate `GenericsStrategy`.

test/com/esotericsoftware/kryo/KryoTestCase.java
MarcMil: I've pulled from master and implemented the changes, so it should work well now.
I am still open for feedback from you and magro :)

theigl: Great thanks! 

Can you please do me a favor and revert all changes not directly related to your PR? It makes it so much harder to review with all the added `@Override` etc. It definitely makes sense to add those but we should do this in a separate PR that doesn't contain any logic changes. Also your import order doesn't seem to match the project import order. Please check this at well.

MarcMil: https://github.com/EsotericSoftware/kryo/blob/master/eclipse/.settings/org.eclipse.jdt.ui.prefs references a formatter profile libgdx. You may want to change that.
You can also enable project-specific code formatter options so that user do not have to do that manually. You can also enable a project-specific code-clean up, which can, among other things, automatically introduce @Overrides to missing methods.
Via Save actions (also in the project settings) these actions can be performed on save automatically, leading to a coherent source code (at least for eclipse users).
You did not specify a particular imports-order. These can be specified in the "Organize Imports" settings page in Eclipse and are saved with an ".importorder" file extension by default. I do not want to edit the imports by hand.

If you are keen on enforcing specific formatting rules, you may consider implementing an approach similar to Soot (https://github.com/Sable/soot), where pull requests are checked for style violations upon merge by an automatic process. I'd highly recommend that and supplying as much automation as possible. Nobody wants to change things like import order by hand...
See https://github.com/Sable/soot/tree/master/codingstyle for a reference on what files you'd needed to supply. Apart from that, as I said, I think preconfiguring the project and committing these rules as well into the repository seems like the easiest way to ensure everyone's on the same page.

I've imported the code-format.xml in Eclipse and ran it over the whole code and commited the result. Some files where not following the code formatting prior either: https://github.com/EsotericSoftware/kryo/pull/683/commits/ef34f1faaf604a8b085915e98c44e47abf795c3d


I'll remove the @Overrides visible in the diff by hand, but honestly, I do not think it is a particular good idea, especially when you plan to introduce them later anyway.

MarcMil:
> So I would suggest that we keep the strategies internal to Kryo and drop `Kryo.setGenerics()` in favor of `Kryo.setOptimizeGenerics(boolean optimizeGenerics)` that internally sets the appropriate `GenericsStrategy`.

Done that. We could also make NoGenericsStrategy and DefaultGenericsStrategy package-visible only, but then these would have to be refactored in order to be relocated from com.esotericsoftware.kryo.util to com.esotericsoftware.kryo.

test/com/esotericsoftware/kryo/KryoTestCase.java
theigl: I fully agree with everything you suggested! I normally use Sonarqube in my projects to enforce code style and quality rules. 

The issue is this: I'm not a developer of Kryo. I'm just a user who likes the project, has been using Kryo 4 for years and wants to see a stable final release of Kryo 5. @magro is the only one currently available that has rights to merge PRs and perform releases. So we have to make it as easy for him as possible to review the PRs and decide whether they can be merged or not. This is why I'm trying to reduce the number of changes in your PR to the absolute minimum necessary to solve the issue at hand.

It would be great if you could help with code style settings and code quality automation after this is merged!

src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -1187,23 +1200,41 @@ public boolean isClosure (Class type) {
 
 	/** Tracks the generic type arguments and actual classes for type variables in the object graph during seralization.
 	 * <p>
-	 * When serializing a type with a single type parameter, {@link Generics#nextGenericClass() nextGenericClass} will return the
-	 * generic class (or null) and must be followed by {@link Generics#popGenericType() popGenericType}. See
-	 * {@link CollectionSerializer} for an example.
+	 * When serializing a type with a single type parameter, {@link DefaultGenericsStrategy#nextGenericClass() nextGenericClass}
```
magro, line 1203: Should this (and the following) rather refer to `GenericsStrategy` instead of `DefaultGenericsStrategy`? Probably this is a result of a rename ;-)

src/com/esotericsoftware/kryo/serializers/AsmField.java
```diff
@@ -23,7 +23,7 @@
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 import com.esotericsoftware.kryo.serializers.FieldSerializer.CachedField;
-import com.esotericsoftware.kryo.util.Generics.GenericType;
+import com.esotericsoftware.kryo.util.DefaultGenericsStrategy.GenericType;
```
magro, line 26: Not sure about this: does it make sense, that `GenericType` is defined within `DefaultGenericsStrategy` (and not within `GenericsStrategy`)? On the one hand I think yes, one the other hand I wonder why some classes should refer to s.th. that's closely related to a specific implementation...

src/com/esotericsoftware/kryo/util/GenericsStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. */
+/** Provides an interface on how generics are handled. */
+public interface GenericsStrategy {
+	/** Sets the type that is currently being serialized. Must be balanced by {@link #popGenericType()}. Between those calls, the
+	 * {@link GenericType#getTypeParameters() type parameters} are returned by {@link #nextGenericTypes()} and
+	 * {@link #nextGenericClass()}. */
+	public void pushGenericType (GenericType fieldType);
```
magro, line 33: Nitpick: this `public` here and below is not needed ... But it can also stay here, I just stumbled over it :-)

src/com/esotericsoftware/kryo/util/NoGenericsStrategy.java
```diff
@@ -0,0 +1,68 @@
...
+ * - Neither the name of Esoteric Software nor the names of its contributors may be used to endorse or promote products derived
+ * from this software without specific prior written permission.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. */
+/** This generic implementation does not infer generics types at all. This leads to a slightly larger serialization output.
```
magro, line 27: Nitpick: a newline above would help to spot this class level comment. I looked for it, but only saw it on a second sight ;-)

test/com/esotericsoftware/kryo/KryoTestCase.java
```diff
@@ -72,25 +74,46 @@ public void setUp () throws Exception {
 		kryo = new Kryo();
 	}
 
+	/** @param length Pass Integer.MIN_VALUE to disable checking the length. */
```
magro, line 77: Nitpick: param comment does not match any of the actual params.

src/com/esotericsoftware/kryo/util/NoGenericsStrategy.java
```diff
@@ -0,0 +1,68 @@
...
+ * However, this implementation is slightly faster and more robust. */
+public final class NoGenericsStrategy implements GenericsStrategy {
```
magro, line 29: I think this could be moved up to the `kryo` package and made package visible only, to reduce the public api.

src/com/esotericsoftware/kryo/util/DefaultGenericsStrategy.java
```diff
@@ -31,9 +30,11 @@
 import java.lang.reflect.TypeVariable;
 import java.util.ArrayList;
 
+import com.esotericsoftware.kryo.Kryo;
+
 /** Stores the generic type arguments and actual classes for type variables in the current location in the object graph.
  * @author Nathan Sweet */
-public class Generics {
+public class DefaultGenericsStrategy implements GenericsStrategy {
```
magro, line 37: I think this could be moved up to the `kryo` package and made package visible only, to reduce the public api.

src/com/esotericsoftware/kryo/util/NoGenericsStrategy.java
theigl: Copyright and legal info should be moved to the very top of the class above the `package` statement as it is done in other files.

test/com/esotericsoftware/kryo/KryoTestCase.java
theigl: @magro: Are you using Eclipse for working on Kryo? We need a working `.importorder` file to avoid imports getting shuffled. 

I'm using IntelliJ with Eclipse formatter and it seems to work fine.

theigl: The current import order seems to be something like this:

```
static all other, 
blank, 
com.*, 
blank, 
java.*, 
blank, 
javax.*, 
blank, 
org.*, 
blank, 
all other imports
```

theigl: Nevermind! I just handcrafted the `.importorder` file: https://github.com/EsotericSoftware/kryo/pull/726/commits/52f272ed8a89d5bbc5c9b01fe2c7eab7501c735f

 I'm not sure how `*` imports are configured in Eclipse and if this setting is included in the current `codeformat.xml`. Static imports always use `*` in the current codebase.

@MarcMil please import it and apply it to your PR.

magro: Great! I'm using intellij, Nate is using Eclipse AFAIK.

src/com/esotericsoftware/kryo/Kryo.java
MarcMil: You're right :)

src/com/esotericsoftware/kryo/serializers/AsmField.java
MarcMil: That was the result of renaming the old Generics class. As you said, it does not make much sense for GenericsType and GenericsHierarchy to reside in the DefaultGenericsStrategy.
Note that since it is still be accessible via the GenericsStrategy, it is still part of the public interface (which it also should be, given that it was already be usable from users of this library).
Instead of having them as static inner classes, I've moved them to their own files. I personally think this is cleaner, but we probably could create them as inner classes of the interface as well. They seem to be data classes and implementation independent as well, so I think it should not be necessary to abstract them using interfaces as well?

src/com/esotericsoftware/kryo/util/GenericsStrategy.java
MarcMil: You mean the modifier of the methods? In ReferenceResolver, the methods are declared public as well, so I figured it'd make sense here as well. I think it looks better as well :)

src/com/esotericsoftware/kryo/util/NoGenericsStrategy.java
MarcMil: You're right, done

src/com/esotericsoftware/kryo/util/NoGenericsStrategy.java
MarcMil: Done

src/com/esotericsoftware/kryo/util/DefaultGenericsStrategy.java
MarcMil: DOne

src/com/esotericsoftware/kryo/util/NoGenericsStrategy.java
theigl: @MarcMil: This class is still public and in the`util` package. Did you push all your local changes?

theigl: Nevermind! Github showed my an old changeset for some reason.

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
```diff
@@ -185,15 +184,19 @@ public void testClassHierarchyWithMultipleTypeVariables () {
...
-		kryo.writeClassAndObject(buffer, o);
+		try {
+			kryo.setOptimizedGenerics(false);
+			kryo.setRegistrationRequired(false);
+			Output buffer = new Output(512, 4048);
+			kryo.writeClassAndObject(buffer, o);
+		} finally {
+			// reset default
+			kryo.setOptimizedGenerics(true);
+		}
```
theigl, line 199: Could you please replace this with `roundTrip(length, o, false)`? I don't know why I didn't use `roundtrip` when I added the test.

test/com/esotericsoftware/kryo/serializers/FieldSerializerTest.java
```diff
@@ -19,7 +19,24 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import org.junit.Test;
+import org.objenesis.strategy.StdInstantiatorStrategy;
```
theigl, line 39: Your import order is still wrong for some files. Please apply the `.importorder` settings to those files.

theigl:
This looks good to me apart from minor comments! 

And this would be great:

> Can you also add a paragraph to the README, @MarcMil? I'd assume that this would really be helpful to users running into issues with generics... Thanks!

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
```diff
@@ -169,8 +169,7 @@ public void testClassHierarchyWithChangingGenericTypeVariables () {
 		ClassHierarchyWithChangingTypeVariableNames.A<?> o = new ClassHierarchyWithChangingTypeVariableNames.A<>(Enum.class);
 
 		kryo.setRegistrationRequired(false);
-
-		roundTrip(131, o);
+		roundTrip(131, o, false);
```
theigl, line 172: Why did you add `false`? This should pass with both strategies.

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
theigl: I just realized that I didn't use roundtrip because the test class does not implement `equals/hashcode`. Please ignore. I'll improve this test later.

theigl:
@MarcMil: Here is a suggestion for a minimal Readme entry:

### Generics

By default, Kryo attempts to use generic type information to optimize for smaller size. If an object's generic type can be inferred, serializers do not need to write the object's class. Generics optimization is enabled or disabled with Kryo `setOptimizedGenerics`.

Disabling generics optimization can increase performance at the cost of a larger serialized size.

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
MarcMil: You're right, removed false and reran test cases

test/com/esotericsoftware/kryo/serializers/FieldSerializerTest.java
MarcMil: Fixed

test/com/esotericsoftware/kryo/serializers/FieldSerializerGenericsTest.java
```diff
@@ -23,12 +23,15 @@
 import com.esotericsoftware.kryo.KryoTestCase;
 import com.esotericsoftware.kryo.io.Output;
 import com.esotericsoftware.kryo.serializers.FieldSerializer.FieldSerializerConfig;
+import com.esotericsoftware.kryo.serializers.FieldSerializerTest.HasGenerics;
+import com.esotericsoftware.kryo.serializers.FieldSerializerTest.ListContainer;
```
theigl, line 27: These two classes are now only used in `FieldSerializerGenericsTest`. Please move them to this class.

src/com/esotericsoftware/kryo/util/GenericsStrategy.java
```diff
@@ -0,0 +1,68 @@
...
+ * from this software without specific prior written permission.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. */
+
+package com.esotericsoftware.kryo.util;
```
theigl, line 20: The interface is still in the `util` package. If the implementations are now in the root package, the interface should be moved there as well.

test/com/esotericsoftware/kryo/serializers/FieldSerializerTest.java
theigl: Thanks!

theigl:
Thanks for your effort @MarcMil! I commented on two more minor things I just noticed.

@magro: Could you please take a second look at this? Imho this looks good now.

theigl:
@magro: I'm not sure moving all these classes to the root package was such a good idea. The package was quite clean before and now has 5 additional classes. 

I would have only moved the interface there, including `Generics` and `GenericsHierarchy` as inner classes and kept the implementations in the `util` package. The same way it is done for `ReferenceResolver`, `ClassResolver` etc. It does not matter that the implementations are public because users cannot do anything with these classes. There is no way to apply a custom `GenericsStrategy` to Kryo at the moment.

I know you suggested this and I can live with it, but it does not seem quite right and is not consistent with the way other interfaces and their implementations are packaged.

MarcMil:
@magro  @theigl 
While there is no way to implement a custom GenericsStrategy for the user, the classes are nevertheless accessible since serializers need to be able to use them.

This is the case already: via ```kryo.getGenerics()``` the user can get a Generics class instance. As such, the user has access to methods like  ```public GenericType[] nextGenericTypes ()``` and ```public int pushTypeVariables (GenericsHierarchy hierarchy, GenericType[] args)```.
As such, GenericType and GenericsHierarchy are already essentially public API I think, so changing anything public in there would break compatibility as well, being inner classes or not.
I personally think that having these classes as separate classes is cleaner than having them as static inner classes when they can (and need to) be referenced by outside classes.

theigl:
> I personally think that having these classes as separate classes is cleaner than having them as static inner classes when they can (and need to) be referenced by outside classes.

That's fine with me. I don't have a strong preference for keeping them as inner classes.

I'm just not sure about the implementation classes being in the root page. The advantage is that we can make them package private, but it isn't consistent with other interface/implementation pairs. But again, I can live with it if you and @magro prefer this solution.

theigl:
I had another walk and another realization. 

As @MarcMil pointed out, the `Generics` class and its subclasses are already *very much public*. There is a whole section in the readme about the class and its related classes. For some reason, I completely ignored that fact until now. The current changes break that API and all custom serializers using it.

We can *easily* stay fully backwards compatible though and I think we should:
- `GenericsStrategy` should be renamed to `Generics` and kept in `util`. 
- `GenericType` and `GenericsHierarchy` should stay inner classes of `Generics`
- The implementations should be renamed to `NoGenerics` and `DefaultGenerics` and either kept in the root package as package private classes or also moved back to the `util` package

@MarcMil: I'm sorry I realized that so late in the progress of this PR! I think we should resolve this issue without breaking the public API. Especially since there is no downside to it and the class names stay as concise as before.

If you have enough of moving classes back and forth, I can also take over and finish your PR.

magro:
Superseded by #734

Issue #684, opened 2019-08-01
Recommendation for CompatibleFieldSerializer


maxisvest:
I read the example in #643, and I see that in @NathanSweet last comment says when decode `SomeClass#anotherClass` but `AnotherClass`'s structure is missing, it will cause decode failure.  Because `CompatibleFieldSerializer` serialize is great depandency in order.  When reading, the `AnotherClass` class doesn't exist, Kryo will skip `AnotherClass` 's chunked input buffer.
See `CompatibleFieldSerializer#read`'s `readUnknownTagData` part down below.

```
Registration registration;
try {
	registration = kryo.readClass(fieldInput);
} catch (KryoException ex) {
	if (!chunked)
		throw new KryoException("Unable to read unknown data (unknown type). (" + getType().getName() + ")", ex);
	if (DEBUG) debug("kryo", "Unable to read unknown data (unknown type).", ex);
		inputChunked.nextChunk();
		continue;
}
```

But the information of `SomeClass#value` is still int that buffer, so if we continue read the remaing buffer that we can cache the `SomeClass#value` in `SomeClass#anotherClass`'s chunked input.
So I simply do that:
```
Registration registration;
try {
	registration = kryo.readClass(fieldInput);
} catch (KryoException ex) {
	if (!chunked)
		throw new KryoException("Unable to read unknown data (unknown type). (" + getType().getName() + ")", ex);
	if (DEBUG) debug("kryo", "Unable to read unknown data (unknown type).", ex);
		//inputChunked.nextChunk();
		//continue;
               registration = kryo.getRegistration(Object.class);
}
```

If Kryo can't find the Class so we give it a default Class that has no field in it, then Kryo successfully read the remaing byte in buffer and cache the `SomeClass#value` in `SomeClass#anotherClass`'s chunked input.

This change is usefull in my case, but I don't know it will have impact for other case? Or you consider add this change (maybe with configurable) in futrue release?

Looking forward to your reply.

NathanSweet:
Sorry for the very late response!

You lost me at:

> But the information of SomeClass#value is still int that buffer,

Each field is written chunked, so when we can't read a field and `inputChunked.nextChunk();`, we only skip the bytes for that field.

Maybe you can explain more how you are using your proposed change?

maxisvest:
@NathanSweet 

> is still **int** that buffer

 is a wrong spelling，I mean 

> But the information of SomeClass#value is still **in** that buffer

[...]

Issue #643, opened 2019-01-27
CompatibleFieldSerializerFactory sometimes doesn't work while add new object as field


liusf1993:
I have a class in server side and client side. And the class is like this:
```
  static public class SomeClass {
    String value;
  }
```
In the next version, there are some change, and the new  structure is like this.
 ```
static public class SomeClass {

    String value;
    AnotherClass anotherClass;
  }

  static public class AnotherClass {

    String k;
    String v;
  }
```
While some value is the same with the another, there is an IndexOutOfBoundsException happened.
Both the code below are tested in the same class, here I separate encode from decode.
Encode:
```
public class CompatibleTest {

  public static final Kryo kryo = new Kryo();

  static {
    CompatibleFieldSerializerFactory compatibleFieldSerializerFactory = new CompatibleFieldSerializerFactory();
    CompatibleFieldSerializerConfig config = compatibleFieldSerializerFactory.getConfig();
    config.setChunkedEncoding(true);
    kryo.setDefaultSerializer(compatibleFieldSerializerFactory);
    kryo.setRegistrationRequired(false);
  }


  @Test
  public void encode() throws IOException {
    String s = "Hello Kryo!";
    AnotherClass anotherClass = new AnotherClass();
    anotherClass.k = s;
    anotherClass.v = s;

    SomeClass someClass = new SomeClass();
    someClass.anotherClass = anotherClass;
    someClass.value = s;

    Output output = new Output(new FileOutputStream("file.bin"));
    kryo.writeObject(output, someClass);
    output.close();

    System.out.println(someClass);
  }

  static public class SomeClass {

    String value;
    AnotherClass anotherClass;
  }


  static public class AnotherClass {

    String k;
    String v;
  }
}
```
Decode:

```
public class CompatibleTest {

  public static final Kryo kryo = new Kryo();

  static {
    CompatibleFieldSerializerFactory compatibleFieldSerializerFactory = new CompatibleFieldSerializerFactory();
    CompatibleFieldSerializerConfig config = compatibleFieldSerializerFactory.getConfig();
    config.setChunkedEncoding(true);
    kryo.setDefaultSerializer(compatibleFieldSerializerFactory);
    kryo.setRegistrationRequired(false);
  }


  @Test
  public void decode() throws IOException {
    Input input = new Input(new FileInputStream("file.bin"));
    SomeClass object2 = kryo.readObject(input, SomeClass.class);
    System.out.println(object2);
    input.close();
  }

[...]

PR #686, opened 2019-08-12
Add a default registration if missing Class structure when re…

maxisvest:
An recommendation for #643 in #684 , give a default registration if missing Class structure when reading

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -135,9 +135,14 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 				} catch (KryoException ex) {
 					String message = "Unable to read unknown data (unknown type). (" + getType().getName() + "#" + cachedField + ")";
 					if (!chunked) throw new KryoException(message, ex);
-					if (DEBUG) debug("kryo", message, ex);
-					inputChunked.nextChunk();
-					continue;
+                    try {
+                        //try to give a default registration
+                        registration = kryo.getRegistration(Object.class);
+                    } catch (Exception e) {
+                        if(DEBUG) trace("Unable to get default registration: " + e.getMessage());
+                        inputChunked.nextChunk();
+                        continue;
+                    }
 				}
 				if (registration == null) {
 					if (chunked) inputChunked.nextChunk();


maxisvest:
@magro , I reopen this PR for trigger CI rebuild and it has been passed, log is here https://travis-ci.org/github/EsotericSoftware/kryo/builds/687970498, and I don't know why this page show that CI build is still failed.

magro:
Not sure either. Maybe rebasing on master would help?

I'd also like to have a test for this, this would also trigger a completely fresh build 😁

src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
```diff
@@ -135,9 +135,14 @@ public T read (Kryo kryo, Input input, Class<? extends T> type) {
 				} catch (KryoException ex) {
 					String message = "Unable to read unknown data (unknown type). (" + getType().getName() + "#" + cachedField + ")";
 					if (!chunked) throw new KryoException(message, ex);
-					if (DEBUG) debug("kryo", message, ex);
-					inputChunked.nextChunk();
-					continue;
+                    try {
```
theigl, line 138: Please correct the indentation of this block.

maxisvest: Done here. It's no difference in my IDEA editor but show wrong format in the page. LOL

theigl:
@maxisvest: If you want to get this merged, please provide a testcase that demonstrates what you are trying to achieve with your change.

maxisvest:
@theigl : Yes，I will write the test case later. But the change I was provide is used to the case of transfer data in two services, and those two services rely one DTO(Data transfer Object) with different version. Which means this test case can't run automaticlly in one service because you should change DTO structure manually to complete the test case.

maxisvest:
@theigl : The test case has been added.

theigl:
@maxisvest: Thank you! I tried your test-case against the current master and it does not throw any exceptions in my case. Could this somehow depend on the JDK vendor and version used? I'm using OpenJDK 11 and the class is deserialized without any issue after deleting the `SubData`.

maxisvest:
@theigl : I think this is not depended on any special JDK version. Just like deserializition on other class. Or should I add a switch for this feature?

theigl:
@maxisvest: We first need a test-case that can reliably replicate the issue. Please merge the current master into your branch and make sure that your test-case really fails. Currently, I cannot reproduce the issue.

On master, my bean looks like this after following all the steps in your test-case:

image

No exception is thrown.

maxisvest:
@theigl : Yes, after merge master the issue is disappeared. Because in `FieldSerializer` will sort `cachedFields` by field name. It will make `message` in front of `subData` and serialize `message` first. So I simplely rename `subData` to `aSubData`. Make sure `aSubData` serialize first in `DataBean`.  And then run test agian in current master, my issue will be reliably replicated.

theigl:
@maxisvest: Thanks. I can now reproduce your issue and I think I understand what's going on: When chunked encoding is enabled and Kryo encounters an unknown class, it skips the whole chunk as described in the JavaDoc:

> In either case, if the data is skipped and {@link Kryo#setReferences(boolean) references} are enabled, then any
> references in the skipped data are not read and further deserialization receive the wrong references and fail.

Your fix works by forcing Kryo to read the chunk to a generic object. It solves the problem but could theoretically cause incorrect behavior for other users. We could disable it by default but I'm not sure what we would call the configuration option.

@magro: Do you have an opinion on this?

magro:
I think I don't understand the consequences completely, e.g. in combination with references.
E.g. I couldn't tell what the documentation for the  new setting would be 😉 Having that would help me. 

So far I have a slight feeling that this change adds more complexity than value, but that's just a feeling.


Another thought:

In my projects I find it quite naturally to do multi step deployments. E.g. if I want to remove a field from a table/db which is used by multiple instances, at first I change the code to no longer use it (deployment 1) and then I remove the column from the table (deployment 2).

Couldn't the same pattern be applied here?
At first add the class so that all parties/services have that class (deployment 1), then use the class in RPCs (deployment 2)?

maxisvest:
@magro : Yes, Theoretically upgrade a common class should take multi step as you says. But to our service system may have special situation on depolyment that is sometime services can't deploy at same time. So it cause the differentiation and I try to avoid it.

PR #705, opened 2019-11-15
Realloc instead of clearing large array list.

johnou:
Right now the Object array inside the readObjects list is never trimmed so if a large object is cloned the JVM will never be able to reclaim the memory.

Screenshot from 2019-11-15 09-38-32

### DIFF
--- a/src/com/esotericsoftware/kryo/util/MapReferenceResolver.java
+++ b/src/com/esotericsoftware/kryo/util/MapReferenceResolver.java
@@ -30,9 +30,21 @@ import java.util.ArrayList;
  * for get or put.
  * @author Nathan Sweet */
 public class MapReferenceResolver implements ReferenceResolver {
+	private static final int CAPACITY = 2048;
+	private static final int MAXIMUM_CAPACITY = 8192;
+
 	protected Kryo kryo;
-	protected final IdentityObjectIntMap writtenObjects = new IdentityObjectIntMap();
-	protected final ArrayList readObjects = new ArrayList();
+	protected final IdentityObjectIntMap<Object> writtenObjects = new IdentityObjectIntMap<>(CAPACITY);
+	protected final ArrayList<Object> readObjects = new ArrayList<>(CAPACITY);
+	private final int maximumCapacity;
+
+	public MapReferenceResolver () {
+		this(MAXIMUM_CAPACITY);
+	}
+
+	public MapReferenceResolver (int maximumCapacity) {
+		this.maximumCapacity = maximumCapacity;
+	}
 
 	public void setKryo (Kryo kryo) {
 		this.kryo = kryo;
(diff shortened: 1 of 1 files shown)


johnou:
@NathanSweet ptal, I also fixed CI.

.travis.yml
```diff
@@ -10,41 +10,15 @@ env:
...
-  # see also https://docs.travis-ci.com/user/languages/java/#using-java-10-and-later
-  # and https://github.com/sormuras/bach#install-jdksh (because default travis docs didn't work)
-  - export JAVA_HOME=$HOME/openjdk11
-  - $TRAVIS_BUILD_DIR/install-jdk.sh --feature 11 --target $JAVA_HOME
-  - mvn -v && mvn -B test
-  # install and publish snapshot with java 8
-  # not via `after_success` but directly in `script`, since a failure would be silently ignored in `after_success`, which
-  # is documented in https://docs.travis-ci.com/user/customizing-the-build#Breaking-the-Build
-  - jdk_switcher use oraclejdk8
-  - if [ $TRAVIS_BRANCH = "master" ] && [ $TRAVIS_PULL_REQUEST = "false" ] && [ $IS_SNAPSHOT ]; then
```
magro: Removing all this is not what we want I'd say.

The build in the past built/compiled with java 8, and then ran tests with java9 and java11, to verify if the published artifact (built with java8) works with java9 and java11.

AFAICS, this is no longer the case with this change.

My approach for this oracle java 8 issue would have been to add
```
dist: trusty
```
following  https://travis-ci.community/t/solved-oraclejdk8-installation-failing-still-again/3428 and https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

Not sure though if this works here

johnou: @magro oraclejdk8 is EOL, swapped to openjdk8 (still receiving updates - LTS), 9 and 10 are eol, no updates, included openjdk11 for tests, by default travis installs the artifacts and runs all unit tests - https://docs.travis-ci.com/user/languages/java#projects-using-maven , then there's an after script hook to deploy a snapshot (built with 8) to sonatype like the previous setup.

johnou: https://travis-ci.org/EsotericSoftware/kryo/builds/612316865?utm_source=github_status&utm_medium=notification shows the results of these travis configuration changes.

johnou: @magro so essentially you'd like me to rollback the jdk_switcher changes to test a java8 binary on different java runtimes?

johnou: @magro ptal I rolled back the travis changes.

NathanSweet:
@magro What do you think of the CI changes? It's greek to me.

magro:
@johnou sorry for the late response, and thanks for your changes! LGTM now. /cc @NathanSweet

johnou:
@magro I rebased with my fixes to Travis CI but it's failing with the new pom-versioned.xml.. 'artifactId' with value 'kryo${kryo.major.version}' does not match a valid id pattern. @ com.esotericsoftware.kryo:kryo${kryo.major.version}:[unknown-version], /home/travis/build/EsotericSoftware/kryo/pom-versioned.xml, line 12, column 14

johnou:
I'll try and spend some time on it this afternoon.

magro:
Awesome, thanks!

magro:
@johnou I now also had a look what's wrong (in master), and noticed that in the versioned pom the parent version was outdated. I just submitted a PR for fixing this, which also includes the switch to openjdk8 and trusty: #713

Hopefully you've not yet spent time on this... ;-)

magro:
The PR #713 passed therefore I merged it to master. Now there are conflicts here in `.travis.yml` - sorry for that!

johnou:
@magro your changeset looks good to me! I rebased the pull request, PTAL.

magro:
Cool, LGTM!

PR #707, opened 2019-12-26
Add test for objects extending others with generics

Tostino:
This is a test which represents a real error I am getting in my main code base since trying to update from Kryo 4.x to Kryo 5 RC5.

### DIFF
--- a/src/com/esotericsoftware/kryo/util/GenericsUtil.java
+++ b/src/com/esotericsoftware/kryo/util/GenericsUtil.java
@@ -108,6 +108,10 @@ public class GenericsUtil {
 			}
 		}
 
+		// We have exhausted looking through superclasses for a concrete generic type
+		// definition, so the current type must have been defined on the first class.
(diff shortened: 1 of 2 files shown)


Tostino:
Working through the issues myself, and will be providing a new pull request ASAP with any fixes I can muster in place.

theigl:
@Tostino: Why did you close this PR? Did your fix turn out to cause other problems?

Tostino:
Yes, this PR caused other problems that it took me a little bit to figure out.

I actually spent about a week working on this, but was unable to figure out how to fix the other corner cases.  So in the end I had to abandon trying to update the library at that point because I don't have the resources to spend there.

theigl:
Thanks for the info @Tostino! Do you still have test-cases for the corner cases you weren't able to fix?

I'm trying to collect and simplify all currently failing test-cases and prepare a PR that can act as a test-suite for fixing generics.

PR #711, opened 2020-02-12
using writeClassAndObject instead of writeObject for captured args in ClosureSerializer#write

ppp-kohe:
As my understanding, ClosureSerializer#write should write class names for captured args by writeClassAndObject instead of writeObject.


NathanSweet:
Cheers!

PR #717, opened 2020-05-16
Test-cases for generics serialization issues

theigl:
This PR collects tests for all currently known issues with generics serialization as suggested in https://github.com/EsotericSoftware/kryo/issues/654#issuecomment-628847087.

I have simplified all test cases to the bare minimum required reproduce the issues.

PR #685, opened 2019-08-12
Add resolve GenericArrayType in GenericsUtil#resolveTypeVariable()

maxisvest:
Try to fix #655 problem, seems like missing an option with `GenericArrayType`

PR #718, opened 2020-05-16
Fixes for generic serialization issues

theigl:
This PR applies known fixes to resolve failing test-cases from #717.

It includes fixes suggested in:
- https://github.com/EsotericSoftware/kryo/pull/685
- https://github.com/EsotericSoftware/kryo/pull/707

All tests pass after applying these changes, but there might be additional corner cases as reported in #707. 

However, I believe that these 3 minor changes will resolve the vast majority of generics issues and should be reasonably safe:

- https://github.com/EsotericSoftware/kryo/commit/aefcde517ddd9ccf6020da52703b1e38c5e138a1 simply adds a missing handler for generic array types
- https://github.com/EsotericSoftware/kryo/commit/f2b6e292cc3f442140d54d3d62b6240004545f36 only applies in cases where an exception is currently thrown and can't really make matters worse
- https://github.com/EsotericSoftware/kryo/commit/ef353ecb0d9805a934e36a782b96d8dc5ee3246f is the only change that could potentially have negative effects on currently working code, but it is simple enough and all tests pass

I would suggest to merge this, create another RC and get more feedback from users.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
@@ -129,6 +129,10 @@ public class FieldSerializer<T> extends Serializer<T> {
 	/** Prepares the type variables for the serialized type. Must be balanced with {@link #popTypeVariables(int)} if >0 is
 	 * returned. */
 	protected int pushTypeVariables () {
+		if (genericsHierarchy.isEmpty()) {
+			return 0;
+		}
+
 		GenericType[] genericTypes = kryo.getGenerics().nextGenericTypes();
 		if (genericTypes == null) return 0;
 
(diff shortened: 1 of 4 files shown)


theigl:
We should also consider #683. It does not change the current generics handling at all but lets users opt-out of generics optimization altogether. 

It could be an alternative to this PR (although it would require users to change their Kryo configuration) or an addition to this PR for users who still encounter corner-cases with generics handling.

@magro: WDYT?

magro:
> We should also consider #683. It does not change the current generics handling at all but lets users opt-out of generics optimization altogether.
> 
> It could be an alternative to this PR (although it would require users to change their Kryo configuration) or an addition to this PR for users who still encounter corner-cases with generics handling.
> 
> @magro: WDYT?

I'm open to this. AFAICS it has merge conflicts - do you want to come up with a cleaned up version of it?

magro:
Awesome, thanks for the PR! I'll publish a new RC...

theigl:
@magro: Thanks a lot!

> I'm open to this. AFAICS it has merge conflicts - do you want to come up with a cleaned up version of it?

I already pinged the owner and left two comments. Let's wait for his feedback.

magro:
RC6 is released and on it's way to maven central...

PR #719, opened 2020-05-17
Fix test when building with JDK11

theigl:
This PR corrects a test failure when building with JDK11.

Locale providers changed with JDK11, resulting in slightly different default behavior:

- https://javachannel.org/posts/help-my-java-locale-is-wrong-in-jdk11/
- https://www.oracle.com/java/technologies/javase/jdk11-suported-locales.html#providers


magro:
Awesome, thanks!

PR #720, opened 2020-05-19
#655 Fix generics corner cases

theigl:
This PR is a follow-up to #718.

The original PR solved many currently known generics issues, but I uncovered two more corner cases while testing against my main app.

The test `ClassHierarchyWithChangingTypeVariableNames` shows that in some cases it is **impossible** to correctly resolve the generic type using the current generics logic. It relies on matching type parameter names, but the test uses different names for the same type in different levels of the inheritance hierarchy.

While this sounds like bad news, this knowledge leads to a very simple solution. Instead of throwing an exception when type parameters cannot be resolved, we simply have to give up and return the type variable.


magro:
Awesome, thanks for your work, @theigl!

Issue #721, opened 2020-05-19
ClassCastException when serializing generic field


theigl:
In same cases generic type resolution can resolve an incorrect type for a field:

```
Caused by: java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String (java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap')
	at com.esotericsoftware.kryo.serializers.DefaultSerializers$StringSerializer.write(DefaultSerializers.java:155)
	at com.esotericsoftware.kryo.Kryo.writeObjectOrNull(Kryo.java:618)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:83)
	... 29 more
```

Test:

```
	@Test
	public void testClassHierarchyWithConflictingTypeVariables () {
		ClassWithConflictingTypeArguments.A o = new ClassWithConflictingTypeArguments.A(
			new ClassWithConflictingTypeArguments.B<>(1));

		kryo.setRegistrationRequired(false);

		Output buffer = new Output(512, 4048);
		kryo.writeClassAndObject(buffer, o);
	}

	static class ClassWithConflictingTypeArguments {
		static final class A {
			C<String> c;

			public A (C<String> c) {
				this.c = c;
			}
		}

		static class B<R, V> implements C<V> {
			R v;

			public B (R v) {
				this.v = v;
			}
		}

		interface C<T> {
		}
	}
```

PR #722, opened 2020-05-19
#721 Add test

theigl:
This PR adds an ignored test to demonstrate #721.

### DIFF
--- a/test/com/esotericsoftware/kryo/serializers/GenericsTest.java
+++ b/test/com/esotericsoftware/kryo/serializers/GenericsTest.java
@@ -163,6 +163,19 @@ public class GenericsTest extends KryoTestCase {
 		kryo.writeClassAndObject(buffer, o);
 	}
 
+	// Test for https://github.com/EsotericSoftware/kryo/issues/721
+	@Test
+	@Ignore("Currently fails")
+	public void testClassHierarchyWithConflictingTypeVariables () {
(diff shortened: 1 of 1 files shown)


test/com/esotericsoftware/kryo/serializers/GenericsTest.java
```diff
@@ -361,4 +374,25 @@ public String toString () {
...
+		static final class A {
+			C<String> c;
+
+			public A (C<String> c) {
+				this.c = c;
+			}
+		}
+
+		static class B<R, V> implements C<V> {
+			R v;
```
magro, line 387: To prevent confusion, this `R` could be renamed from `v` to `r`.

```suggestion
			R r;
```

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
```diff
@@ -361,4 +374,25 @@ public String toString () {
...
+
+			public B (R v) {
+				this.v = v;
```
magro, line 390:
```suggestion
			public B (R r) {
				this.r = r;
```

test/com/esotericsoftware/kryo/serializers/GenericsTest.java
theigl: Good catch. Renamed.

PR #723, opened 2020-05-20
#721 Add missing import

theigl:
@magro: For some reason, there was a missing import in the PR you merged yesterday.

Shouldn't Travis also run for PRs and report broken builds?


magro:
Good catch, thanks!

Not sure why it wasn't built, and apparently this one here was not built as well. I couldn't find the reason quickly, will have an eye on it or investigate later.

magro:
Weird: the broken build is shown at travis: https://travis-ci.org/github/EsotericSoftware/kryo/builds/688936998
But it's not shown here...

Issue #693, opened 2019-09-10
Support for java.util.ImmutableCollections


efwe:
With Java9 we got new factory methods like `Set.of`,`List.of` and `Map.of` which use the (non public) class `java.util.ImmutableCollections`. Without a special serializer I get the following error:
```
com.esotericsoftware.kryo.KryoException: java.lang.UnsupportedOperationException
Serialization trace:
longList (de.fw.KryoTest$Immutables)

	at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:144)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.read(FieldSerializer.java:543)
	at com.esotericsoftware.kryo.Kryo.readObject(Kryo.java:709)
	at de.fw.KryoTest.testSerializeJava9ImmutableCollections(KryoTest.java:54)
Caused by: java.lang.UnsupportedOperationException
	at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:134)
	at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:40)
	at com.esotericsoftware.kryo.Kryo.readObject(Kryo.java:731)
	at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:125)
	... 3 more
```

When I configure Kryo like this 
`kryo.register(List.of().getClass(), new JavaSerializer());`
it works. 

My questions are: 
* Do any special serializers already exist?
* Should they exist? 

So this is half question, half bug report here.
Keep up the good work,
Frank

theigl:
I just ran into this issue as well. There should definitely be serializers for these new immutable collections. They can probably be implemented like the immutable collection serializers for Guava in https://github.com/magro/kryo-serializers.

See for instance https://github.com/magro/kryo-serializers/blob/master/src/main/java/de/javakaffee/kryoserializers/guava/ImmutableListSerializer.java

theigl:
Here is a quick implementation of serializers for `java.util.ImmutableCollections`: https://gist.github.com/theigl/7979208a36170b40060ec0c9c4690104

theigl:
@magro: Are you open to creating a JDK11 version of kryo-serializers? If so, I could create a PR with my serializers for `java.util.ImmutableCollections`.

magro:
@theigl can you please submit an issue for kryo-serializers for this? Then I hope to have a look at this next week...

[...]

PR #724, opened 2020-05-20
#693 JDK 11 Build

theigl:
This PR adds support for building with JDK11 while still being able to run tests with JDK8.

This enables Kryo to support JDK9+ classes like `java.util.ImmutableCollections` while still targeting JDK8 environments.

The same approach was taken to support JDK8 classes such as `Optional` (#362). 

This time things are a little more complicated though. JDK9 introduced new covariant return types for some methods on `ByteBuffer`, making the code backwards incompatible (see https://github.com/EsotericSoftware/kryo/issues/693#issuecomment-630096985).

The only way around this issue that does not require additional artifacts for newer JDK versions is to use implicit or explicit casts for these methods. This PR adds these casts. It doesn't look great and is brittle (we have to rely on CI builds to catch accidental new usages of incompatible methods). But it still might make sense to go down this road to support newer JDK classes with minimal effort.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ script:
   # install and publish snapshot with java 8
   # not via `after_success` but directly in `script`, since a failure would be silently ignored in `after_success`, which
   # is documented in https://docs.travis-ci.com/user/customizing-the-build#Breaking-the-Build
-  - jdk_switcher use openjdk8
   - if [ $TRAVIS_BRANCH = "master" ] && [ $TRAVIS_PULL_REQUEST = "false" ] && [ $IS_SNAPSHOT ]; then
       mvn -v && mvn -B -P requireSnapshot -DskipTests=true deploy --settings build/settings-sonatype.xml;
     else
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,32 +15,19 @@ env:
     - secure: kCOLgEveEDprSL3fOBeCc1/nmHVYEcNh/lRRWiJywFehOzKBTBF40pAWX7JtsWUPG8aPKGYFEE4sjYd+DF2DRknbuB2ZfAEgs9Xp9h5Mx1KztJYNiCfEnLciBBRDXA7t1M3TLKyT3ptA+Kk307cH2QHv9sUUnhnf0IofJyQhQvA=
   
 jdk:
-  #- openjdk7
-  - openjdk8
-
-before_install:
-  # required for testing jdk 11
-  - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh -O $TRAVIS_BUILD_DIR/install-jdk.sh
-  - chmod u+x $TRAVIS_BUILD_DIR/install-jdk.sh
+  - openjdk11
 
 install:
-  # compile and test java 8
+  # compile and test java 11
   - mvn -B -V clean test
 
 script:
-  # compile tests with java 8
-  - jdk_switcher use openjdk8
+  # compile tests with java 11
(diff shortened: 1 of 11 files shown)


magro:
Thanks for the PR, great! :-)

I think we could have chosen java 9 as "separator" as well, because this brought classes like `ImmutableCollections`. Probably you preferred java 11 because it's an LTS release - the choice is not important to me and I think we could change it anyways if needed :-)

magro:
The thing that I'm missing here is the adjustment of the `.travis-ci.yml` - do you want to submit this with the next PR? Right now the compilation is done with java 8, which should be java 11 now, right?

theigl:
I updated the `.travis-cli.yml`. I also seems to have fixed the Travis integration somehow. I am now seeing build checks again.

.travis.yml
```diff
@@ -15,32 +15,24 @@ env:
...
 
 before_install:
   # required for testing jdk 11
   - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh -O $TRAVIS_BUILD_DIR/install-jdk.sh
   - chmod u+x $TRAVIS_BUILD_DIR/install-jdk.sh
 
 install:
-  # compile and test java 8
+  # compile and test java 11
   - mvn -B -V clean test
```
theigl, line 27: @magro: Why do we run the tests in the `install` phase? They are run again immediately in the `script` phase by `mvn -B install`. Imho the `install` phase can be removed.

theigl:
> I think we could have chosen java 9 as "separator" as well, because this brought classes like ImmutableCollections. Probably you preferred java 11 because it's an LTS release - the choice is not important to me and I think we could change it anyways if needed :-)

Correct. I chose 11 because 9 and 10 basically do not exist out in the wild anymore. We can always change that if needed.

.travis.yml
magro: Not sure TBH, so I don't know a good reason why we should not get rid of the install stage.

.travis.yml
```diff
@@ -15,32 +15,19 @@ env:
...
-  # test java 9 (only available via oraclejdk9)
-  - jdk_switcher use oraclejdk9
-  - mvn -v && mvn -B test
-  # test java 11
-  # see also https://docs.travis-ci.com/user/languages/java/#using-java-10-and-later
-  # and https://github.com/sormuras/bach#install-jdksh (because default travis docs didn't work)
-  - export JAVA_HOME=$HOME/openjdk11
-  - $TRAVIS_BUILD_DIR/install-jdk.sh --feature 11 --target $JAVA_HOME
+  # test java 8
+  - jdk_switcher use openjdk8
```
magro, line 30: The line "jdk_switcher use openjdk8" can now be removed.

theigl: Done

.travis.yml
theigl: @magro: OK I get it now. The [default behavior](https://docs.travis-ci.com/user/languages/java/#projects-using-maven) of Travis is to run:

```
mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```

It probably was necessary to override this behavior before but I don't see a reason for it now. I have further simplified the build config to be more aligned with Travis default behavior in https://github.com/EsotericSoftware/kryo/commit/8766e270d01760cf3a64efba26381a28ff54dae4 but you merged my PR seconds before I pushed the change.

If you want you can apply the changes manually. Currently, we run one more `install` than is really necessary.

magro: Can you submit a separate PR please?

theigl: Sure! Here you go #728

PR #727, opened 2020-05-21
#693 JDK11 Immutable Collections

theigl:
This PR builds upon #724 to ad support for `java.util.ImmutableCollections`.

The implementation is modelled after the serializers for immutable collections in Guava provided by `kryo-serializers` (https://github.com/magro/kryo-serializers/tree/master/src/main/java/de/javakaffee/kryoserializers/guava)


magro:
Great, thanks!

theigl:
@magro: You are too fast! ;) There was a tiny omission in this PR so I created #729.

Also I was thinking: Does it make sense to optimize for any of these cases: `Map1`, `Set12`, `List12`? Theoretically, we could store those in a more compact way. The `Map1` case is basically the same as `CollectionsSingletonMapSerializer`. The other two cases are more complicated and I'm not sure if its worth optimizing them.

magro:
Thanks for catching this and for the new PR! I should check more carefully as it seems 😉

Regarding optimizations for Map1, Set12, List12: as long as optimizations are understood and maintainable they're a good thing, because efficiency is one major goal ok kryo.

After having a 2nd look at the ImmutableCollectionsSerializer I'm wondering if serialization of Map should share some code with existing MapSerializer instead of creating a new HashMap. For larger maps the creation of the new HashMap could take "significant" time. WDYT?

theigl:
> Regarding optimizations for Map1, Set12, List12: as long as optimizations are understood and maintainable they're a good thing, because efficiency is one major goal of kryo.

I will look into this. At least `Map1` should be straight forward.

> After having a 2nd look at the ImmutableCollectionsSerializer I'm wondering if serialization of Map should share some code with existing MapSerializer instead of creating a new HashMap, because for larger maps the creation of the new HashMap could take "significant" time. WDYT?

That makes sense. The problem is how `MapSerializer` currently constructs maps: It first instantiates a new map, then reads key-values one by one and puts them into the map. This approach cannot work with immutable maps. Maybe I can extract something like a `MapFactory` from the code that gets passed all the deserialized entries. I'll think about this tomorrow.

magro:
Cool!
Just to be clear: I was talking about the implementation of `write`, where a new HashMap is created.

theigl:
> Just to be clear: I was talking about the implementation of `write`, where a new HashMap is created.

I know. The issue is that if I want to extend `CollectionSerializer` and `MapSerializer` for these collections (which would make sense), I have to solve the problem that both of these classes assume that collections they create are *mutable*. This only matters for `read` and `copy` but it is currently impossible to extend from these classes without solving this issue as well. 

The existing `ArraysAsListSerializer` has exactly the same issue: https://github.com/EsotericSoftware/kryo/blob/ff75f80664f0218ea1c681990c2e59e64cc28b27/src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java#L813. 

It extends from `CollectionSerializer` and doesn't have to create an additional list on `write`. But it is forced to create an additional list on `read` because the superclass only supports the mutable "create then add` way of creating collections.

theigl:
@magro: I created #731 that extends the new serializers from existing base classes.

I decided against optimizations for `Map1`, `Set12`, and `List12`. These are implementation details that are intentionally not public. Also the existing serializers for maps and collections do not optimize for cases with only a single element. I think that consistency and simplicity is more important in this case than saving a couple of bytes.

magro:
Cool, I can follow that reasoning.

theigl:
@magro: I sent you a PN on Gitter. Please take a look when you find the time.

Daanielvb:
in which version was the support for immutable collections released @theigl ? I am working with spring state machine and it seems like we're tied with an old version. Do you know if its possible to upgrade kryo version without changing in SSM?

theigl:
@Daanielvb: This was released in Kryo 5.0.0 and I'm currently not planning on backporting it to Kryo 4 because this version targets JDK <= 8.

I would recommend that you upgrade to Kryo 5. The main challenge is that data serialized with Kryo 4 cannot be read with Kryo 5. Probably the easiest way for you to move forward without breaking anything is to create another artifact (`spring-statemachine-kryo5`) that depends on the [versioned artifact](https://github.com/EsotericSoftware/kryo#with-maven) of Kryo 5 and let your users chose which version they want to use.

PR #728, opened 2020-05-21
#693 Align `travis.yml` with default Travis behavior

theigl:
This PR uses the same approach as the [default Travis build](https://docs.travis-ci.com/user/languages/java/#projects-using-maven) for Maven projects:

- The `install` phase runs `mvn install`
- The `script` phase runs `mvn test`

The only difference is that I dropped `-Dmaven.javadoc.skip=true` from the `install` phase so JavaDocs are still generated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,14 @@ env:
 jdk:
   - openjdk11
 
+install:
+  # build with java 11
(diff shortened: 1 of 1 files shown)


magro:
Great, thanks!

PR #729, opened 2020-05-21
#693 Ensure copied map is immutable

theigl:
I missed this in #727. The returned map should be immutable as well.


magro:
Good catch, and I missed that as well! 😏

PR #731, opened 2020-05-22
#693 Extend existing serializers for immutable collections

theigl:
The serializers for `java.util.ImmutableCollections` added in #727 were adapted from the Guava serializers in the `kryo-serializers` project. They did not re-use the existing `CollectionSerializer` and `MapSerializer` classes that are heavily optimized for binary size.

This PR adjusts the new serializers to extend the existing base classes. 

The advantages of this approach are:

- It is consistent with the existing `ArraysAsListSerializer` that also extends `CollectionSerializer`
- It produces a consistent and optimized binary representation
- It does not allocate an additional `HashMap` when writing maps

There is a downside though:

All serializers now allocate an additional list/set/map on `read` and `copy` because the base classes [do not support](https://github.com/EsotericSoftware/kryo/pull/727#issuecomment-632330971) creating immutable objects.

However, it is more important to get the binary representation right so we do not have to break serialization compatibility in the future. The internal logic of the serializers can be improved in a follow-up PR.


magro:
Cool, LGTM!

PR #733, opened 2020-05-30
#721 Only store type variables if all of them have type arguments

### LINKED ISSUE #721: ClassCastException when serializing generic field

theigl:
This PR resolves #721. It does not store generic type parameters if there aren't as many arguments as type parameters at the root of the generics hierarchy.

This is the cleanest solution I could think of and the previously failing test-case for #721 is the only test affected by the new condition. However, it is only a workaround. It might actually be possible to match the argument to the correct type parameter, but given that this is a rare corner case, I think we can live with not optimizing generics.

### DIFF
--- a/src/com/esotericsoftware/kryo/util/Generics.java
+++ b/src/com/esotericsoftware/kryo/util/Generics.java
@@ -117,6 +117,9 @@ public class Generics {
 	 * @param args May contain null for type arguments that aren't known.
 	 * @return The number of entries that were pushed. */
 	public int pushTypeVariables (GenericsHierarchy hierarchy, GenericType[] args) {
+		// Do not store type variables if we do not have arguments for all of them
+		if (args.length < hierarchy.rootTotal) return 0;
+
 		int startSize = this.argumentsSize;
 
 		// Ensure arguments capacity.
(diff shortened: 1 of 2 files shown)


theigl:
@magro: Please review. I hope this is the last of the generics corner cases.

magro:
Thanks for the PR, LGTM!

PR #738, opened 2020-07-06
Consolidate generics hierarchy checks

theigl:
This PR moves the check in `FieldSerializer` introduced in #654 into `DefaultGenerics` so all hierarchy verification is done in that class.

This is purely a cleanup PR and should not change any behavior.

@magro: Please take a look.


magro:
Great, LGTM!

PR #739, opened 2020-07-06
Add missing `@Override` annotations

theigl:
#683 showed that it is difficult to maintain the current code style for contributors. Two issues with the code base are the most problematic:

1. Missing `@Override` annotations on overridden methods
2. Non-standard sorting of modifiers (`static public final` instead of the standard `public static final`)

Both issues will be highlighted as warnings in IDEs and are most likely fixed by default cleanup actions. 

When adding new methods and classes, contributors have to remember these idiosyncrasies and manually sort modifiers and remove override annotations to match the current style.

I suggest that we address both issues before releasing Kryo 5 to make future contributions easier.

This PR addresses the first issue and adds all missing `@Override` annotations. The second issue is addressed in #740.

@magro: Please take a look.


magro:
Awesome, thanks!

PR #745, opened 2020-07-28
#744 CompatibleFieldSerializer and fields declared as super-types

### LINKED ISSUE #744: CompatibleSerializer Issue when upgraded to 5.0
Currently we are using kryo version: "com.esotericsoftware:kryo:5.0.0-RC4" but when I upgraded to "com.esotericsoftware:kryo:5.0.0-RC5" our test cases are failing. These is our serializer configs:
```
kryoPool = new Pool<Kryo>(true, true) {
      protected Kryo create() {
        Kryo kryo = new Kryo();
        kryo.setReferences(false);
        kryo.setCopyReferences(false);
        kryo.setInstantiatorStrategy(new StdInstantiatorStrategy());
        CompatibleFieldSerializer.CompatibleFieldSerializerConfig config = new CompatibleFieldSerializer.CompatibleFieldSerializerConfig();
        config.setChunkedEncoding(true);
        config.setReadUnknownFieldData(true);
        kryo.setDefaultSerializer(new SerializerFactory.CompatibleFieldSerializerFactory(config));
        return kryo;
      }
```

We are using compatibleSerializer and I saw that to resolve one of the issue there was a change made in CompatibleSerializer's write method(**link**: https://github.com/EsotericSoftware/kryo/commit/4ea2a8a3acd99a1f499d5439b8033d8678bae3cf). I think this change is failing test cases. In version it used to get specific class but now it takes type which is Java.lang.Object for us. So while writing nested object using compatibleSerializer it's not able to register the nested object in CacheFields and not able to encode it.
This is the object we want to encode and decode:
```
public class CacheObject {

  @JsonFormat(shape = JsonFormat.Shape.STRING)
  private Date stored;
  private long ttl;
  private Object value;

  public CacheObject() { }

  public CacheObject(Date stored, long ttl, Object value) {
    this.stored = stored;
    this.ttl = ttl;
    this.value = value;
  }

  public Date getStored() {
    return stored;
  }

  public long getTtl() {
    return ttl;
  }

  public Object getValue() {
    return value;
  }
}
```
For the object which is inside "value" attribute it's not able to initialize CacheFields and that's why not able to write that object.

theigl:
This PR attempts to resolve #744.

The original fix pushed by @NathanSweet in https://github.com/EsotericSoftware/kryo/commit/fb10f37ffec43d461b10d0cf4f17b950d12b6752 only works for fields that have the same type as the object stored in it. It fails for fields declared as super-types and interfaces.

This PR adds a check for primitive types originally proposed by @isaki in #670 and uses the concrete value class in all other instances.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -90,7 +90,13 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 				try {
 					if (object != null) {
 						Object value = cachedField.field.get(object);
-						if (value != null) valueClass = cachedField.field.getType();
+						if (value != null) {
+							final Class<?> fieldType = cachedField.field.getType();
+							if (fieldType.isPrimitive())
+								valueClass = fieldType;
+							else
+								valueClass = value.getClass();
+						}
 					}
 				} catch (IllegalAccessException ex) {
 				}
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -30,6 +30,7 @@ import com.esotericsoftware.kryo.io.InputChunked;
 import com.esotericsoftware.kryo.io.Output;
 import com.esotericsoftware.kryo.io.OutputChunked;
 import com.esotericsoftware.kryo.util.ObjectMap;
+import com.esotericsoftware.kryo.util.Util;
 
 /** Serializes objects using direct field assignment, providing both forward and backward compatibility. This means fields can be
  * added or removed without invalidating previously serialized bytes. Renaming or changing the type of a field is not supported.
@@ -90,13 +91,7 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 				try {
 					if (object != null) {
 						Object value = cachedField.field.get(object);
-						if (value != null) {
-							final Class<?> fieldType = cachedField.field.getType();
-							if (fieldType.isPrimitive())
-								valueClass = fieldType;
-							else
-								valueClass = value.getClass();
-						}
+						if (value != null) valueClass = value.getClass();
 					}
 				} catch (IllegalAccessException ex) {
 				}
@@ -168,7 +163,7 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 				}
 
 				// Ensure the type in the data is compatible with the field type.
-				if (cachedField.valueClass != null && !cachedField.valueClass.isAssignableFrom(valueClass)) {
+				if (cachedField.valueClass != null && !Util.isAssignableTo(valueClass, cachedField.valueClass)) {
 					String message = "Read type is incompatible with the field type: " + className(valueClass) + " -> "
 						+ className(cachedField.valueClass) + " (" + getType().getName() + "#" + cachedField + ")";
 					if (!chunked) throw new KryoException(message);
--- a/src/com/esotericsoftware/kryo/util/Util.java
+++ b/src/com/esotericsoftware/kryo/util/Util.java
@@ -27,6 +27,8 @@ import com.esotericsoftware.kryo.serializers.FieldSerializer;
 import com.esotericsoftware.kryo.util.Generics.GenericType;
 
 import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
 
 /** A few utility methods, mostly for private use.
  * @author Nathan Sweet */
@@ -52,6 +54,18 @@ public class Util {
 
 	// Maximum reasonable array length. See: https://stackoverflow.com/questions/3038392/do-java-arrays-have-a-maximum-size
 	public static final int maxArraySize = Integer.MAX_VALUE - 8;
+
+	private static final Map<Class<?>, Class<?>> primitiveWrappers = new HashMap<>();
+	static {
+		primitiveWrappers.put(boolean.class, Boolean.class);
+		primitiveWrappers.put(byte.class, Byte.class);
+		primitiveWrappers.put(char.class, Character.class);
+		primitiveWrappers.put(double.class, Double.class);
+		primitiveWrappers.put(float.class, Float.class);
+		primitiveWrappers.put(int.class, Integer.class);
+		primitiveWrappers.put(long.class, Long.class);
+		primitiveWrappers.put(short.class, Short.class);
+	}
 
 	public static boolean isClassAvailable (String className) {
 		try {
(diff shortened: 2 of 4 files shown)


theigl:
@isaki: I'm not 100% sure my solution achieves the same as yours. You mentioned a failing unit test due to primitive boxing in #670. Was it an existing test or one you added locally to verify your fix?

isaki:
It was something failing in my test applications. I basically had multiple JARs built with different implementations of the same class to trigger the JVM crash. My final PR had all the required fixes.

theigl:
Thanks for the info @isaki!

My implementation also prevents boxing, but it still sets the field's `valueClass`. I have to think about the potential consequences some more. If I'm not 100% sure that my solution works in all cases, I'll replace it with the conditional from your original PR tomorrow.

theigl:
I was able to reproduce the test failures you mentioned @isaki. I never ran the tests before I added my condition to prevent boxing. When I remove the condition, a bunch of tests fail with "Read type is incompatible with the field type: int -> Integer" etc. 

So I guess we are good to go and can apply this PR.

@magro: Please take a look.

isaki:
I actually think this might run into issues; in my original investigation setting value class even to the primitive type caused problems. Based on the "Files changed" tab you are assigning valueClass in all cases and then invoking the set method on the cachedField instance. I suspect that you have reintroduced the issues my checks intended to prevent. I could be wrong; it's been awhile since I looked at this.

Unfortunately I no longer have the project that I used to generate these issues, however, it would not take me long to recreate. I shall endeavor to do so later today/tomorrow. I will test both my approach and yours to see if the things work correctly. If it turns out my paranoia wasn't required, all the better!

theigl:
> Unfortunately I no longer have the project that I used to generate these issues, however, it would not take me long to recreate. I shall endeavor to do so later today/tomorrow. I will test both my approach and yours to see if the things work correctly. If it turns out my paranoia wasn't required, all the better!

That would be great! We can add the condition from your PR, but I would feel much more comfortable doing it because of a failing test-case.

theigl:
> I actually think this might run into issues; in my original investigation setting value class even to the primitive type caused problems. Based on the "Files changed" tab you are assigning valueClass in all cases and then invoking the set method on the cachedField instance. I suspect that you have reintroduced the issues my checks intended to prevent. I could be wrong; it's been awhile since I looked at this.

It is possible that this causes other issues, but not very likely because the fix applied by Nate also sets the `valueClass` in *all* cases. Do you remember if you verified your tests against Nate's fixes? If so, we should be relatively safe.

isaki:
I did not test against Nate's fixes, no; I had assumed he had.

isaki:
This is very time consuming, but the current release version has at least one bug. My test setup involved commenting in and out various lines to create 3 different jars. One used a `String` field, one used a `Long` field, and one used a `long` field; all three had the same class and name.

I've pushed my work in progress code to here: https://github.com/isaki/kryo-jvm-test

## Kryo 5.0.0-RC7 Release ##

| Source | Target | Result | Notes |
|:------:|:------:|:------:|:-----:|
| String | String | Success | N/A |
| String | Long   | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: String -> Long (io.isaki.kryojvmtest.Payload#id) | N/A |
| String | long   | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: String -> long (io.isaki.kryojvmtest.Payload#id) | N/A |
| Long   | String | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: long -> String (io.isaki.kryojvmtest.Payload#id) | N/A |
| Long   | Long   | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: long -> Long (io.isaki.kryojvmtest.Payload#id) | BUG, likely in serialization given the exception text. |
| Long   | long   | io.isaki.kryojvmtest.Payload@210366b4 (19000) | The primitive based JAR read the boxed value no problem |
| long   | String | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: long -> String (io.isaki.kryojvmtest.Payload#id) | N/A |
| long   | Long   | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: long -> Long (io.isaki.kryojvmtest.Payload#id) | This could probably be made to work. |
| long   | long   | io.isaki.kryojvmtest.Payload@210366b4 (18000) | The primitive based JAR read the primitive value no problem |

I believe that the current implementation is broken in that it is serializing a boxed value as a primitive. I double checked my build and ensured I was working with a boxed value.

theigl:
@isaki: Thank you so much for your work! 

I can reproduce the bug you found, but I don't fully understand it yet. The value is clearly written as a `Long` but read as a `long` for some reason. The same thing happens on my branch and the condition from your PR does not work either.

I think we have to add a check for primitives and their wrapper types instead of simply checking if the classes are assignable. This would also support changes from `long` to `Long` and vice versa.

theigl:
I just pushed a better solution for this issue in https://github.com/EsotericSoftware/kryo/pull/745/commits/bbe0acaa2885cbd45c00c3a0641c1577e0f901e7

I completely removed the condition that checks for primitive classes when serializing and made the assignable check aware of primitive types and their wrappers. I also added a test that should guard against these issues in the future.

isaki:
I'll build your patch and run my tests later this afternoon.

isaki:
My testing results show things look good except for one small case.

| Source | Target | Result | Notes |
|:------:|:------:|:------:|:-----:|
| String | String | io.isaki.kryojvmtest.Payload@67784306 (stringid) | N/A |
| String | Long   | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: String -> Long (io.isaki.kryojvmtest.Payload#id) | N/A |
| String | long   | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: String -> long (io.isaki.kryojvmtest.Payload#id) | N/A |
| Long   | String | com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: long -> String (io.isaki.kryojvmtest.Payload#id) | BUG? That should read Long - > String |
| Long   | Long   | io.isaki.kryojvmtest.Payload@eec5a4a (19000) | N/A |
| Long   | long   | io.isaki.kryojvmtest.Payload@67784306 (19000) | Nice! |
| long   | String |com.esotericsoftware.kryo.KryoException: Read type is incompatible with the field type: long -> String (io.isaki.kryojvmtest.Payload#id)|N/A|
| long   | Long   |io.isaki.kryojvmtest.Payload@eec5a4a (18000)|Nice!|
| long   | long   |io.isaki.kryojvmtest.Payload@67784306 (18000)|N/A|

theigl:
@isaki: Thanks again! I would consider the remaining issue more a problem of wording than a real bug. I don't think it has any consequences other than a slightly incorrect error message.

@magro: Please review. This fixes #744 and resolves a number of hidden issues with CompatibleFieldSerializer.

jgiannuzzi:
Thanks a lot for your work on fixing these issues, @theigl and @isaki!

We are also hitting #744 and it prevents us from upgrading to the latest Kryo, even though it contains other fixes that we need. Could you please review this PR, @magro and @NathanSweet? Thanks!

magro:
Many thanks for your really great work here, @theigl and @isaki!

PR #750, opened 2020-07-31
#746 Ensure pool maximum capacity is honored

### LINKED ISSUE #746: IllegalStateException on calling free for pooled item to synchronized pool using soft references and being full
Usig com.esotericsoftware.kryo.util.Pool with threadSafe = true and softReferences = true and a limit.

On calling free and the pool already being full, an IllegalStateException is thrown: 
`java.lang.IllegalStateException: Queue full
	at java.util.AbstractQueue.add(AbstractQueue.java:98) ~[?:?]
	at com.esotericsoftware.kryo.util.Pool$SoftReferenceQueue.offer(Pool.java:157) ~[76:com.esotericsoftware.kryo:5.0.0.RC5]
	at com.esotericsoftware.kryo.util.Pool.free(Pool.java:86) ~[76:com.esotericsoftware.kryo:5.0.0.RC5]`

A LinkedBlockingQueue is used in this case with the "add" method not bein overwritten and being called in the SoftReferenceQueue's offer method.

This doe not happen with one of the boolean parameters set to false.

theigl:
This PR resolves #746.

Limiting a pools maximum capacity is currently broken for two cases:

1. threadSafe=true softReferences=true
2. threadSafe=false softReferences=false

In the first case, `LinkedBlockingQueue.add` isn't overridden, so an `IllegalStateException` is thrown when reaching maximum capacity.

In the second case, the wrong method was overridden. `ArrayDeck` is not wrapped in a `SoftReferenceQueue` so `ArrayDeck.offer` is called directly. Internally, it uses `addLast` instead of `add` so the overridden method is never called.

I also made a slight change to the code formatting settings that preserves explicit linebreaks. It greatly improves the readability of the test-cases.

### DIFF
--- a/eclipse/code-format.xml
+++ b/eclipse/code-format.xml
@@ -163,7 +163,7 @@
 <setting id="org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode" value="enabled"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_before_closing_paren_in_switch" value="do not insert"/>
 <setting id="org.eclipse.jdt.core.formatter.alignment_for_method_declaration" value="0"/>
-<setting id="org.eclipse.jdt.core.formatter.join_wrapped_lines" value="true"/>
+<setting id="org.eclipse.jdt.core.formatter.join_wrapped_lines" value="false"/>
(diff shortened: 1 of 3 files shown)


theigl:
@magro: Please review.

magro:
Cool, thanks for the PR!

PR #755, opened 2020-08-10
Use CuckooObjectMap from Kryo 4 for class registrations

theigl:
This PR adds back the `CuckooObjectMap` from Kryo 4 and uses it for mapping class registrations.

The new map implementations in Kryo 5 are ideal for read/write workloads. Nearly all maps in Kryo are heavily written to before they are cleared after each round of serialization. So this is a perfect fit.

There is one exception though: The map for class registrations in `DefaultClassResolver`. This map is never cleared and read much more often than written.

This is a use-case that the new maps do not handle well. Kryo 4's implementation of `ObjectMap` has significantly higher read throughput. This becomes especially noticeable when you serialize a large number of different classes. For maps with 1000-3000 entries, CuckooObjectMap has 40-60% faster reads:

**Kryo 4 ObjectMap**

Benchmark  |  (numItems)  |   Mode |        Score   | Units
------------ | ------------- | ------------- |   --: | -------------
MapBenchmark.read   |        100 | thrpt   | 887534,878 |   ops/s
MapBenchmark.read   |       1000 | thrpt   | **172467,301** |   ops/s
MapBenchmark.read   |       3000 | thrpt   | **150534,662** |   ops/s
MapBenchmark.read   |      10000 | thrpt   |  66231,572 |   ops/s

**Kryo 5 ObjectMap**

Benchmark  |  (numItems)  |   Mode |        Score   | Units
------------ | ------------- | ------------- |   --: | -------------
MapBenchmark.read   |        100 | thrpt   | 727130,171 |   ops/s
MapBenchmark.read   |       1000 | thrpt   | **122685,225** |   ops/s
MapBenchmark.read   |       3000 | thrpt   |  **92703,749** |   ops/s
MapBenchmark.read   |      10000 | thrpt   |  63471,399 |   ops/s

**HashMap** (for reference)

Benchmark  |  (numItems)  |   Mode |        Score   | Units
------------ | ------------- | ------------- |  --: | -------------
MapBenchmark.read   |        100 | thrpt  | 772165,893      |   ops/s
MapBenchmark.read   |       1000 | thrpt  | **137372,335**  |   ops/s
MapBenchmark.read   |       3000 | thrpt  | **122497,382** |   ops/s
MapBenchmark.read   |      10000 | thrpt  |  62414,886 |   ops/s

Benchmark Implementation: The benchmark setup writes `numItems` classes generated with ByteBuddy to the map. The benchmark itself then retrieves `Math.min(numItems, 500)` randomly selected classes back from the map.

### DIFF
--- a/src/com/esotericsoftware/kryo/util/CuckooObjectMap.java
+++ b/src/com/esotericsoftware/kryo/util/CuckooObjectMap.java
@@ -1,0 +1,820 @@
+/* Copyright (c) 2008-2020, Nathan Sweet
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
+ * conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the distribution.
+ * - Neither the name of Esoteric Software nor the names of its contributors may be used to endorse or promote products derived
+ * from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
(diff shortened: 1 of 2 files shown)


theigl:
@magro: Please review.

magro:
Awesome, thanks!

tommyettinger:
Is the benchmark available anywhere? It might be handy for me; I've been trying to see if there's any way to improve upon the system used in Kryo 5 (and libGDX) by maybe avoiding `long` math in the Fibonacci hashing. I don't know where the read bottleneck might be, but it seems to go away at 10K items. That seems like there might be an issue with how much of the capacity is currently used, since 10K is above 8192, the previous table size, but 1K and 3K are closer to 1024 and 4096, respectively, and probably just passed the load factor's threshold. I'm guessing cuckoo hashing will just be as fast as it generally is with high-quality hashcodes, and since a Class uses the identity hashcode, that should be fine re: collisions. There's some JVM flags you can tinker with if you want to make Class.hashCode() use a different algorithm, for testing: `-XX:+UnlockExperimentalVMOptions -XX:hashCode=2` I would expect to fail catastrophically (it makes the identity hashCode() always return 1), but if you change the `2` at the end to a different option, [there's some documentation on that here](https://alidg.me/blog/2020/7/15/hash-code).

theigl:
@tommyettinger: My benchmark looked like this: https://gist.github.com/theigl/6da09a3873059c25d828dada55e14e77

I used classes generated via ByteBuddy to be as close to the real use case of the map in `DefaultClassResolver`, but you can easily replace that with other types of objects and get rid of the ByteBuddy dependency.

If you drop the file into `kryo-benchmarks`, you should be able to run it.

tommyettinger:
@theigl Thank you very much! I hadn't checked back for a reply and I was half-awake-thinking about other ways I could come close to the same conditions, but your ByteBuddy approach seems like the best way to replicate the exact situation with `Class` instances as keys. If the only requirement was to use the same hashCode() and equals() as `Class`, then I could just use `Object`, though the memory used per key and value would be different, and I don't know if that would affect caching.

theigl:
@tommyettinger: We just received a PR #845 for improving the performance of read-mostly ObjectMaps. The benchmark numbers look good, but I'd like to get feedback by someone who has a better understanding of the map implementations. Could you take a look at the PR?

PR #758, opened 2020-08-12
#754 Unimplement `Poolable` from Kryo

### LINKED ISSUE #754: Kryo is reset twice when used in a pool
Kryo 5 has a `Poolable` interface for objects that can be used with `Pool`.

When such an object is returned to the pool, `Poolable.reset()` is called:

	/** Called when an object is freed to clear the state of the object for possible later reuse. The default implementation calls
	 * {@link Poolable#reset()} if the object is {@link Poolable}. */
	protected void reset (T object) {
		if (object instanceof Poolable) ((Poolable)object).reset();
	}

This makes using `Input` und `Output` much easier in combination with pools.

The problem is that the `Kryo` class implements `Poolable` as well. So if `autoReset` is enabled (the default), `Kryo.reset()` will be called twice. First by auto-reset at the end of serialization, then when the object is returned to the pool.

This operation is potentially expensive, especially when references are enabled and class registration is not required.

We have two options to fix this:

1. Do not let `Kryo` implement `Poolable`
2. Document that `autoReset` should be disabled when Kryo is used in a pool

@magro: What do you think? We should address this before releasing 5.0 Final.

theigl:
This PR removes `Poolable` from Kryo to prevent resetting the object graph state twice after each (de)serialization. 

`Poolable` is a good fit for `Input`/`Output`, but in the case of `Kryo` conflicts with the concept of `autoReset`.

Resolves #754.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1236,7 +1236,7 @@ If `true` is passed as the second argument to the Pool constructor, the Pool sto
 
 The third Pool parameter is the maximum capacity. If an object is freed and the pool already contains the maximum number of free objects, the specified object is reset but not added to the pool. The maximum capacity may be omitted for no limit.
 
-If an object implements Pool.Poolable then Poolable `reset` is called when the object is freed. This gives the object a chance to reset its state for reuse in the future. Alternatively, Pool `reset` can be overridden to reset objects. Input and Output implement Poolable to set their `position` and `total` to 0. Kryo implements Poolable to reset its object graph state.
+If an object implements Pool.Poolable then Poolable `reset` is called when the object is freed. This gives the object a chance to reset its state for reuse in the future. Alternatively, Pool `reset` can be overridden to reset objects. Input and Output implement Poolable to set their `position` and `total` to 0.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1236,7 +1236,7 @@ If `true` is passed as the second argument to the Pool constructor, the Pool sto
 
 The third Pool parameter is the maximum capacity. If an object is freed and the pool already contains the maximum number of free objects, the specified object is reset but not added to the pool. The maximum capacity may be omitted for no limit.
 
-If an object implements Pool.Poolable then Poolable `reset` is called when the object is freed. This gives the object a chance to reset its state for reuse in the future. Alternatively, Pool `reset` can be overridden to reset objects. Input and Output implement Poolable to set their `position` and `total` to 0. Kryo implements Poolable to reset its object graph state.
+If an object implements Pool.Poolable then Poolable `reset` is called when the object is freed. This gives the object a chance to reset its state for reuse in the future. Alternatively, Pool `reset` can be overridden to reset objects. Input and Output implement Poolable to set their `position` and `total` to 0.
```
magro, line 1239: I thought it would make sense to mention that for a pooled Kryo instance with autoReset=false reset must be called manually before returning it to the pool. WDYT?

theigl: @magro: I just added this to the Readme.

Issue #699, opened 2019-10-08
Problems with boolean serialization/deserialization


esalsbury:
Using 5.0.0-RC4 version.  Here is the basic pool setup:

```
private final Pool<Kryo> kryoPool = new Pool<Kryo>(true, true) {

			@Override
			public Kryo create() {
				Kryo kryo = new KryoReflectionFactorySupport();
				kryo.setDefaultSerializer(new SerializerFactory.BaseSerializerFactory() {
					@Override
					public Serializer newSerializer(Kryo kryo, Class type) {
						CompatibleFieldSerializer.CompatibleFieldSerializerConfig config = new CompatibleFieldSerializer.CompatibleFieldSerializerConfig();
						config.setChunkedEncoding(true);
						config.setReadUnknownFieldData(true);
						CompatibleFieldSerializer serializer = new CompatibleFieldSerializer(kryo, type, config);
						return serializer;
					}
				});
				kryo.setRegistrationRequired(false);
				kryo.setInstantiatorStrategy(new DefaultInstantiatorStrategy(new SerializingInstantiatorStrategy()));

IdValidatingKryoRegistrar registrar = new IdValidatingKryoRegistrar(kryo, RESERVED_KRYO_IDS);
				for (KryoInitializer initializer : initializers) {
					initializer.initialize(registrar);
				}

				return kryo;
			}
```
Serialize method:

```
public byte[] serialize(final Object object) throws SerializationException {
		Kryo kryo = kryoPool.obtain();
		try {
			ByteArrayOutputStream baos = new ByteArrayOutputStream(512);
			Output output = new Output(baos);
			kryo.writeObject(output, object);
			output.close();

			return baos.toByteArray();
		} finally {
			kryoPool.free(kryo);
		}
	}
```
Deserialize method:

```
	public <T> T deserialize(final byte[] bytes, Class<T> clazz) throws DeserializationException {
		Kryo kryo = kryoPool.obtain();
		try {
			ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
			Input input = new Input(bais);
			T object = kryo.readObject(input, clazz);
			input.close();
			return object;
		} finally {
			kryoPool.free(kryo);
		}
	}
```

I can reliably cause a failure with the following test where I serialize and immediately deserialize an object:

```
	@Test
	public void testBrokenObjectABunchOfTimes() {
		Foo foo = new Foo();
		foo.setAccountId(-8490936757185012249L);
[...]

PR #759, opened 2020-08-19
[kryo-699] Add an exception for extendedFieldName

Mr14huashao:
reporduche issue #699 
Add an exception for the same FieldName


theigl:
Thank you for this PR @Mr14huashao!

I would prefer a solution that does not allocate on every `write` operation. We could override [`initializeCachedFields`](https://github.com/EsotericSoftware/kryo/blob/78f33328c35a5affbd577465b6dae2cea73b3010/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java#L87) for `CompatibleFieldSerializer` and do the check there. This method is called only once per class. 

A warning that includes which fields are duplicates is probably better than an exception.

Does this really only affect `CompatibleFieldSerializer` or does `FieldSerializer` have the same problem?

PR #763, opened 2020-08-29
Add an exception for extendedFieldName 

Mr14huashao:
Add a test case about boolean field for #699

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -31,6 +31,7 @@ import com.esotericsoftware.kryo.io.Output;
 import com.esotericsoftware.kryo.io.OutputChunked;
 import com.esotericsoftware.kryo.util.ObjectMap;
 import com.esotericsoftware.kryo.util.Util;
+import java.util.ArrayList;
 
 /** Serializes objects using direct field assignment, providing both forward and backward compatibility. This means fields can be
  * added or removed without invalidating previously serialized bytes. Renaming or changing the type of a field is not supported.
@@ -64,11 +65,18 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 
 		CachedField[] fields = cachedFields.fields;
 		ObjectMap context = kryo.getGraphContext();
+		ArrayList fieldsList = new ArrayList();
 		if (!context.containsKey(this)) {
 			if (TRACE) trace("kryo", "Write fields for class: " + type.getName());
 			context.put(this, null);
 			output.writeVarInt(fields.length, true);
 			for (int i = 0, n = fields.length; i < n; i++) {
+				fieldsList.add(fields[i].name);
+				if (i + 1 < n) {
+					if (fieldsList.contains(fields[i+1].name))
+						throw new IllegalArgumentException(fields[i].field.getDeclaringClass() +
+								" has the same field as the super class, config.ExtendedFieldNames must be true");
+				}
 				if (TRACE) trace("kryo", "Write field name: " + fields[i].name + pos(output.position()));
 				output.writeString(fields[i].name);
 			}
--- a/test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
+++ b/test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
@@ -24,17 +24,28 @@ import static org.junit.Assert.*;
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.KryoException;
 import com.esotericsoftware.kryo.KryoTestCase;
+import com.esotericsoftware.kryo.Serializer;
+import com.esotericsoftware.kryo.SerializerFactory;
 import com.esotericsoftware.kryo.SerializerFactory.CompatibleFieldSerializerFactory;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import com.esotericsoftware.kryo.util.DefaultInstantiatorStrategy;
 import org.apache.commons.lang.builder.EqualsBuilder;
+import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+import org.objenesis.strategy.SerializingInstantiatorStrategy;
 
 /** @author Nathan Sweet */
 public class CompatibleFieldSerializerTest extends KryoTestCase {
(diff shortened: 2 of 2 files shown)


theigl:
@Mr14huashao: As I said in #759, we need a solution with minimal overhead. Ideally, the duplicate field check would be done only once when the serializer is first initialized. I think we should only log a warning so we do not break existing applications and force them to use `extendedFieldNames`.

src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
```diff
@@ -64,11 +65,18 @@ public void write (Kryo kryo, Output output, T object) {
 
 		CachedField[] fields = cachedFields.fields;
 		ObjectMap context = kryo.getGraphContext();
+		ArrayList fieldsList = new ArrayList();
```
theigl, line 68: This declaration should be moved into `if (!context.containsKey(this))`

src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
```diff
@@ -64,11 +65,18 @@ public void write (Kryo kryo, Output output, T object) {
...
 		ObjectMap context = kryo.getGraphContext();
+		ArrayList fieldsList = new ArrayList();
 		if (!context.containsKey(this)) {
 			if (TRACE) trace("kryo", "Write fields for class: " + type.getName());
 			context.put(this, null);
 			output.writeVarInt(fields.length, true);
 			for (int i = 0, n = fields.length; i < n; i++) {
+				fieldsList.add(fields[i].name);
+				if (i + 1 < n) {
+					if (fieldsList.contains(fields[i+1].name))
```
theigl, line 76: This is quite inefficient. Why not use a `HashSet`? `HashSet.add()` returns `false` if the value is already present, so we can do the check in a single operation.

src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
```diff
@@ -64,11 +65,18 @@ public void write (Kryo kryo, Output output, T object) {
...
+						throw new IllegalArgumentException(fields[i].field.getDeclaringClass() +
+								" has the same field as the super class, config.ExtendedFieldNames must be true");
```
theigl, line 78: Since fields are sorted by name, I'm not sure this message will be correct in all cases. I'd change this to

> Duplicate field detected in class hierarchy

test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
```diff
@@ -642,4 +653,117 @@ public int hashCode () {
...
+					final CompatibleFieldSerializer serializer = new CompatibleFieldSerializer(kryo, type, config);
+					return serializer;
+				}
+			});
+			kryo.setRegistrationRequired(false);
+			kryo.setInstantiatorStrategy(new DefaultInstantiatorStrategy(new SerializingInstantiatorStrategy()));
+			return kryo;
+		}
+
+		private class Child extends Father {
```
theigl, line 706: Please make the classes `static` and use more generic class and field names. Something like `ClassWithDuplicateField`, `SuperClass`, `SuperSuperClass`.

theigl (changes requested):

Mr14huashao:
@theigl Thank you for checking my code.
Based on the suggestion you provided, I modified some code, and the code is better.
I've benefited a lot from it. 
Thank you very much.

theigl (changes requested):
@Mr14huashao: Sorry for the delay. I just reviewed your PR.

Please adjust your tests to use existing support from `KryoTestCase` and change your tests to look like others in the same file.

src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
```diff
@@ -58,6 +61,22 @@ public CompatibleFieldSerializer (Kryo kryo, Class type, CompatibleFieldSerializ
 		this.config = config;
 	}
 
+	@Override
+	protected void initializeCachedFields() {
+		if (!super.config.extendedFieldNames) {
```
theigl, line 66: Do you really need to call `super`? Why can't we use the `config` field from this class?

test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
```diff
@@ -642,4 +653,35 @@ public int hashCode () {
...
+		serializer.getCompatibleFieldSerializerConfig().setChunkedEncoding(true);
+		serializer.getCompatibleFieldSerializerConfig().setExtendedFieldNames(false);
+		serializer.updateFields();
+		kryo.register(ClassWithDuplicateField.class, serializer);
+
+		final ClassWithDuplicateField duplicateField = new ClassWithDuplicateField();
+		roundTrip(31, duplicateField);
+
+		ClassWithDuplicateField object2 = (ClassWithDuplicateField) kryo.readClassAndObject(input);
+		assertEquals(duplicateField, object2);
```
theigl, line 670: This `equals` does not make much sense because it does not compare on the `customNote` field. You should include the field in the `equals` contract. What happens then? Does the test fail? 

As I mentioned before, we should also assert that a warning is actually logged. Please see `WarnUnregisteredClassesTest` for how this can be done.

src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
Mr14huashao: config.extendedFieldNames report:
java.lang.NullPointerException at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.initializeCachedFields(CompatibleFieldSerializer.java:66)
	at com.esotericsoftware.kryo.serializers.CachedFields.rebuild(CachedFields.java:111)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.<init>(FieldSerializer.java:82)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.<init>(CompatibleFieldSerializer.java:60)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.<init>(CompatibleFieldSerializer.java:56)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializerTest.testLogWarningOnDuplicateFieldInClassHierarchy(CompatibleFieldSerializerTest.java:666)

Unless setExtendedFieldNames is override in CompatibleFieldSerializerConfig.
If you have a better way, please teach me . Thank you in advance.

test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
Mr14huashao: Agree！Done.

src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
Mr14huashao: @theigl You are right! Done.

theigl:
@Mr14huashao: I'm currently trying to stabilize Kryo 5.0.0 and will merge only bug fix PRs in the near future. Let's revisit this after Kryo 5.0.1 has been released.

Mr14huashao:
@theigl Do I want to modify some content based on the current version?

Issue #764, opened 2020-08-29
Dose kryo need to add test cases?


Mr14huashao:
I found that kryo still had some methods that didn't have test cases.

need add test cases?

theigl:
@Mr14huashao: Test coverage is currently around 80%, which is OK but not too great. I'm planning to add automated coverage reports for CI builds using Codecov so we can monitor and improve the test coverage.

If you want to add some tests, please go ahead and create a PR!

PR #765, opened 2020-08-31
add some test case for kryo

Mr14huashao:
add some test case for Factory, Registration and Copy. #764


Mr14huashao:
update and add some test case.
increase overall coverage by 2%.

Mr14huashao:
The added test case must be in a standard format.
Open it again when the code is complete

PR #769, opened 2020-09-27
Add `formatter-maven-plugin` to easily apply Eclipse code format

theigl:
This PR adds the `formatter-maven-plugin` that allows users of any IDE to easily apply the project code format via `mvn formatter:format`.

It also bumps some plugin dependency versions and adds some plugin declarations to get rid of warnings in IntelliJ.

As a next step, project formatting could be validated during the build by adding the `formatter:validate` goal.


magro (approved):
Great, LGTM!

PR #772, opened 2020-10-10
Fix explicit KryoSerializer to have higher prio than implicit choices.

### LINKED ISSUE #771: Explicit Serializer directives should be used.
If you have a class that implementes/extends two classes/interfaces for which a predefined DefaulSerializer is available the priority is always a point for discussion.
However the KryoSerializable is one which is to be considered an "explicit" serializer choice.
At this moment if you have a class that does 
`class MyCustomMap extends TreeMap<String, String> implements KryoSerializable {`
then for some reason the serializer for the TreeMap 'wins' which I consider to be bad.

See https://github.com/nielsbasjes/BugReport-KryoMap

nielsbasjes:
This is my first attempt to fix #771
Note that I'm not sure if there are any side effects of changing the ordering in the list of defaults.
So please review this thoroughly. 

I have included a set of tests that verifies all scenarios I could think of in this area.

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -193,7 +193,6 @@ public class Kryo {
 		addDefaultSerializer(boolean[].class, BooleanArraySerializer.class);
 		addDefaultSerializer(String[].class, StringArraySerializer.class);
 		addDefaultSerializer(Object[].class, ObjectArraySerializer.class);
-		addDefaultSerializer(KryoSerializable.class, KryoSerializableSerializer.class);
 		addDefaultSerializer(BigInteger.class, BigIntegerSerializer.class);
 		addDefaultSerializer(BigDecimal.class, BigDecimalSerializer.class);
 		addDefaultSerializer(Class.class, ClassSerializer.class);
@@ -222,6 +221,7 @@ public class Kryo {
 		addDefaultSerializer(void.class, new VoidSerializer());
 		addDefaultSerializer(PriorityQueue.class, new PriorityQueueSerializer());
 		addDefaultSerializer(BitSet.class, new BitSetSerializer());
+		addDefaultSerializer(KryoSerializable.class, KryoSerializableSerializer.class);
 		OptionalSerializers.addDefaultSerializers(this);
 		TimeSerializers.addDefaultSerializers(this);
 		ImmutableCollectionsSerializers.addDefaultSerializers(this);
--- a/test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
+++ b/test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
@@ -1,0 +1,122 @@
+/* Copyright (c) 2008-2020, Nathan Sweet
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
+ * conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the distribution.
+ * - Neither the name of Esoteric Software nor the names of its contributors may be used to endorse or promote products derived
+ * from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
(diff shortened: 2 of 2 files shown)


src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -374,6 +374,10 @@ public Serializer getDefaultSerializer (Class type) {
 		Serializer serializerForAnnotation = getDefaultSerializerForAnnotatedType(type);
 		if (serializerForAnnotation != null) return serializerForAnnotation;
 
+		if (type.isAssignableFrom(KryoSerializable.class) && type != Object.class) {
+			return new KryoSerializableSerializer();
+		}
```
theigl, line 379: Why do we need this condition? Shouldn't the subsequent loop return a `KryoSerializableSerializer` anyway? And why the `type != Object.class`?

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Niels Basjes
+ */
+@SuppressWarnings("synthetic-access")
+public class SerializerPriorityTest {
+
+    // Dummy serializer
+    public static final class MyAnnotationSerializer extends Serializer<Object> {
```
theigl, line 43: Please make all the inner classes and interfaces package private. There is no need for them to be public and it makes the code slightly more readable.

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @DefaultSerializer(MyAnnotationSerializer.class)
+    public static final class MyAnnotatedTreeMap extends TreeMap<String, String> {    }
+
+    @DefaultSerializer(MyAnnotationSerializer.class)
+    public static final class MyAnnotatedKryoSerializable implements BaseKryoSerializable {    }
+
+    @DefaultSerializer(MyAnnotationSerializer.class)
+    public static final class MyAnnotatedTreeMapKryoSerializable extends TreeMap<String, String> implements BaseKryoSerializable {    }
+
+    private void checkCorrectSerializer(Class<?> myClass, Class<? extends Serializer<?>> expectedSerializer) {
```
theigl, line 81: Please move this method to the end of the test class. I'd prefer the name `assertSerializer`

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Test
+    public void testKryoSerializable() {
+        checkCorrectSerializer(MyKryoSerializable.class, KryoSerializableSerializer.class);
+    }
+
+    @Test
+    public void testTreeMapKryoSerializable() {
+        // We choose that the KryoSerializable is more important than the default Serializer
```
theigl, line 99: Instead of comments you could also adjust the test names. Something like `testKryoSerializableSerializerHasHigherPriorityThanDefaultSerializer`. This doesn't look great but its much easier to see what's going on when the test fails.

theigl (changes requested):
@nielsbasjes: Thank you for the contribution! I left some minor comments on the test code and I don't fully understand the added condition in `getDefaultSerializer`.

src/com/esotericsoftware/kryo/Kryo.java
nielsbasjes: Yes, you're right. This was wrong.

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
nielsbasjes: Done

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
nielsbasjes: Done

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
nielsbasjes: Done.

src/com/esotericsoftware/kryo/Kryo.java
nielsbasjes: I simplified the fix by only moving the KryoSerializable to the end of the list.

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
theigl: Please remove the comments now that the test names reflect the test purpose.

Apart from that LGTM.

theigl:
> Note that I'm not sure if there are any side effects of changing the ordering in the list of defaults. So please review this thoroughly.

This should only cause side effects when someone implements `KryoSerializable` but expects the default serializer to be used. So the exact opposite of the case you are describing. Since that doesn't make a lot of sense, I think your changes are alright.

test/com/esotericsoftware/kryo/serializers/SerializerPriorityTest.java
nielsbasjes: Done

theigl (approved):

magro:
Great, thanks for the PR!

PR #779, opened 2020-10-21
Upgrade of dependencies

isaki:
Pick up the latest and greatest object instantiation dependencies.

ReflectASM 1.11.8 -> 1.11.9
Objenesis 3.0.1 -> 3.1

### DIFF
--- a/pom-main.xml
+++ b/pom-main.xml
@@ -18,12 +18,12 @@
 		<dependency>
 			<groupId>com.esotericsoftware</groupId>
 			<artifactId>reflectasm</artifactId>
-			<version>1.11.8</version>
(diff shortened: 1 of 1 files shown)


theigl:
Thanks @isaki!

@magro: Is there any reason Kryo hasn't been updated to the latest ReflectASM?

isaki:
I'm overriding both Objenesis and ReflectASM in my dependency management to the versions above with Kryo 5 (JVM is Corretto 11, compiler is Corretto 11, maven build output version is Java 11) without issue, so it may have just been an oversight when pushing 5.0 out the door.

theigl (approved):

PR #788, opened 2020-11-02
#774 Do not re-use serializers for CompatibleFieldSerializer and TaggedFieldSerializer

### LINKED ISSUE #774: CompatibleFieldSerializer fails when superclass fields concrete type changes
It looks like CompatibleFieldSerializer is caching the first concrete type for a field, If you try to pass a different concrete type the serializer fails.

I wrote the following code against "com.esotericsoftware:kryo:5.0.0-RC9"

It basically writes a wrapper instance with a int value, then tries to write another wrapper instance with a string value and fails.

```
public class KryoSerializerDemo {

  public static void main(String[] args) {

    Kryo kryo = new Kryo();

    CompatibleFieldSerializer.CompatibleFieldSerializerConfig config = new CompatibleFieldSerializer.CompatibleFieldSerializerConfig();
    kryo.setDefaultSerializer(new SerializerFactory.CompatibleFieldSerializerFactory(config));

    Output output = new Output(4096, Integer.MAX_VALUE);
    kryo.writeClassAndObject(output, new Wrapper(123));
    output.reset();
    kryo.writeClassAndObject(output, new Wrapper("foo"));
  }

  public static class Wrapper {
    public Wrapper(Object value) {
      this.value = value;
    }

    Object value;
  }
}
```

When I run the code above, I get the following exception:

```
Exception in thread "main" com.esotericsoftware.kryo.KryoException: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
Serialization trace:
value (com.foo.KryoSerializerDemo$Wrapper)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:96)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.write(CompatibleFieldSerializer.java:107)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:640)
	at com.foo.KryoSerializerDemo.main(KryoSerializerDemo.java:26)
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer

	at com.esotericsoftware.kryo.serializers.DefaultSerializers$IntSerializer.write(DefaultSerializers.java:124)
	at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:571)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:86)
```

### LINKED ISSUE #784: CompatibleFieldSerializer throws IllegalArgumentException
I assume the underlying problem is same as in: https://github.com/EsotericSoftware/kryo/issues/774, but just to double check.

Version used is `5.0.0`.

When we run:
```Java
class KryoSerializerDemo {
    public static void main(String[] args) {
        Kryo kryo = new Kryo();

        CompatibleFieldSerializer.CompatibleFieldSerializerConfig config = new CompatibleFieldSerializer.CompatibleFieldSerializerConfig();
        kryo.setDefaultSerializer(new SerializerFactory.CompatibleFieldSerializerFactory(config));

        Output output = new Output(4096, Integer.MAX_VALUE);
        kryo.writeClassAndObject(output, new Car(new Electric()));
        output.reset();
        kryo.writeClassAndObject(output, new Car(new Petrol()));
    }

    static class Car {
        private Motor motor;

        public Car(Motor motor) {
            this.motor = motor;
        }
    }

    interface Motor {

    }

    static class Electric implements Motor {
        private String range = "371km";
    }

    static class Petrol implements Motor {
        private int horsePower = 115;
    }
}
```

The output is:
```
Exception in thread "main" com.esotericsoftware.kryo.KryoException: java.lang.IllegalArgumentException: Can not set java.lang.String field KryoSerializerDemo$Electric.range to KryoSerializerDemo$Petrol
Serialization trace:
motor (KryoSerializerDemo$Car)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:96)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.write(CompatibleFieldSerializer.java:107)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:644)
	at KryoSerializerDemo.main(KryoSerializerDemo.java:17)
Caused by: java.lang.IllegalArgumentException: Can not set java.lang.String field KryoSerializerDemo$Electric.range to KryoSerializerDemo$Petrol
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
Caused by: java.lang.IllegalArgumentException: Can not set java.lang.String field KryoSerializerDemo$Electric.range to KryoSerializerDemo$Petrol

	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.ensureObj(UnsafeFieldAccessorImpl.java:58)
	at java.base/jdk.internal.reflect.UnsafeObjectFieldAccessorImpl.get(UnsafeObjectFieldAccessorImpl.java:36)
	at java.base/java.lang.reflect.Field.get(Field.java:419)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.write(CompatibleFieldSerializer.java:93)
	at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:575)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:86)
	... 3 more
```

After some debugging, the problem seems to be that when writing `Car` class with `Petrol` engine (second call), cached fields from first write call are returned [here](https://github.com/EsotericSoftware/kryo/blob/35294de1b17e03d4ebb6f440ae6e24cea6967440/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java#L65).

theigl:
Resolves #774 
Resolves #784

This PR prevents re-using `ReflectField` serializers when using `CompatibleFieldSerializer` and `TaggedFieldSerializer`. 

Re-using serializers works for `FieldSerializer` because it only provides a concrete `valueClass` for final field types or if `fixedFieldTypes` is enabled. `CompatibleFieldSerializer ` and `TaggedFieldSerializer` *always* populate the `valueClass` with the last value they encounter.

I'm not fond of this solution but it is the only way to resolve this issue without breaking backwards compatibility.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -242,6 +242,11 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 		return fields;
 	}
 
+	@Override
+	boolean supportsReuse() {
+		return !config.readUnknownFieldData;
+	}
+
 	public CompatibleFieldSerializerConfig getCompatibleFieldSerializerConfig () {
 		return config;
 	}
--- a/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializer.java
@@ -47,7 +47,7 @@ import com.esotericsoftware.kryo.util.Util;
 public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 	private static final int binarySearchThreshold = 32;
 
-	private CompatibleFieldSerializerConfig config;
+	private final CompatibleFieldSerializerConfig config;
 
 	public CompatibleFieldSerializer (Kryo kryo, Class type) {
 		this(kryo, type, new CompatibleFieldSerializerConfig());
@@ -102,6 +102,7 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 				}
 				cachedField.setCanBeNull(false);
 				cachedField.setValueClass(valueClass);
+				cachedField.setReuseSerializer(false);
 			}
 
 			cachedField.write(fieldOutput, object);
@@ -174,6 +175,7 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 
 				cachedField.setCanBeNull(false);
 				cachedField.setValueClass(valueClass);
+				cachedField.setReuseSerializer(false);
 			} else if (cachedField == null) {
 				if (!chunked) throw new KryoException("Unknown field. (" + getType().getName() + ")");
 				if (TRACE) trace("kryo", "Skip unknown field.");
@@ -240,11 +242,6 @@ public class CompatibleFieldSerializer<T> extends FieldSerializer<T> {
 
 		kryo.getGraphContext().put(this, fields);
 		return fields;
-	}
-
-	@Override
-	boolean supportsReuse() {
-		return !config.readUnknownFieldData;
 	}
 
 	public CompatibleFieldSerializerConfig getCompatibleFieldSerializerConfig () {
--- a/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java
@@ -211,11 +211,6 @@ public class FieldSerializer<T> extends Serializer<T> {
 		return (T)kryo.newInstance(original.getClass());
 	}
 
-	/** If true, the serializer supports re-use of serializers for all instances of a field */
-	boolean supportsReuse() {
-		return true;
-	}
-
 	@Override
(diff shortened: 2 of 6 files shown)


theigl:
@magro: If you have a couple of minutes to spare, please take a look at this PR. This fix should make it into Kryo 5.0.1.

I asked Nate for feedback why we always set the `valueClass` for `Compatible` and `Tagged` serializers but I haven't heard from him yet. Removing the `valueClass` from those serializers would resolve this issue as well, but it would break serialization compatibility. So I guess this solution will have to do for now and we can revisit this for Kryo 6.

Issue #804, opened 2021-01-07
llegalStateException on calling free for pooled item to synchronized pool using soft references and being full


wyhasany:
Using com.esotericsoftware.kryo.util.Pool with threadSafe = true and softReferences = false and a limit.

```
java.lang.IllegalStateException: Queue full
	at java.base/java.util.AbstractQueue.add(AbstractQueue.java:98)
	at com.esotericsoftware.kryo.util.Pool$1.add(Pool.java:52)
	at com.esotericsoftware.kryo.util.Pool$SoftReferenceQueue.offer(Pool.java:168)
	at com.esotericsoftware.kryo.util.Pool.free(Pool.java:95)
	at serializers.KryoSnappyRedisSerializer.serialize(KryoSnappyRedisSerializer.java:79)
```

I'm running following code under JMH with 8 concurrent threads. Is that correct behavior?
```java
package tech.viacomcbs.serializers;

import com.esotericsoftware.kryo.Kryo;
import com.esotericsoftware.kryo.SerializerFactory;
import com.esotericsoftware.kryo.io.Input;
import com.esotericsoftware.kryo.io.Output;
import com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer;
import com.esotericsoftware.kryo.util.DefaultInstantiatorStrategy;
import com.esotericsoftware.kryo.util.Pool;
import org.objenesis.strategy.StdInstantiatorStrategy;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.data.redis.serializer.RedisSerializer;
import org.springframework.data.redis.serializer.SerializationException;

import java.io.ByteArrayOutputStream;
import java.io.IOException;

public class KryoRedisSerializer implements RedisSerializer<Object> {

    public static final Logger log = LoggerFactory.getLogger(KryoRedisSerializer.class);

    static final byte[] EMPTY_ARRAY = new byte[0];

    private static final Pool<Kryo> kryoPool = new Pool<Kryo>(true, true, 16) {
        @Override
        protected Kryo create() {
            Kryo kryo = new Kryo();
            try {
                kryo.setRegistrationRequired(false);
                CompatibleFieldSerializer.CompatibleFieldSerializerConfig config = new CompatibleFieldSerializer.CompatibleFieldSerializerConfig();
                config.setExtendedFieldNames(false);
                config.setReadUnknownFieldData(true);
[...]

PR #805, opened 2021-01-07
Prevent ISE Queue full #804

wyhasany:
Prevent throwing exception on multi-threaded environment where two
concurrent threads can change state between these lines:
```
if (size() >= maximumCapacity) return false;
super.add(o);
```
what causes mentioned exception.


theigl:
Good catch @wyhasany! Thanks you!

wyhasany:
@theigl Thanks for merging 👍

PR #826, opened 2021-05-05
Upgrade Maven Bundle Plugin to 5.1.2 to support jdk15

0roman:
otherwise you will run into ConcurrentModificationException such as in https://github.com/FasterXML/oss-parent/issues/27

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,7 +190,7 @@
 				<plugin>
 					<groupId>org.apache.felix</groupId>
 					<artifactId>maven-bundle-plugin</artifactId>
-					<version>4.2.1</version>
+					<version>5.1.2</version>
 				</plugin>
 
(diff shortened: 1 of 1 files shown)


theigl:
@0roman: Thanks for the PR! 

The bundle plugin had a bug that prevented us from upgrading to it:

- https://issues.apache.org/jira/browse/FELIX-6312
- https://issues.apache.org/jira/browse/FELIX-6337

This seems to have been fixed in 5.1.2, so the upgrade should be fine now.

We currently override the plugin version for JDK16 here: https://github.com/EsotericSoftware/kryo/blob/master/pom.xml#L423-L428

This override is unnecessary now. Please remove it, then I can merge your PR.

theigl:
Merged, thanks!

PR #831, opened 2021-05-18
Allowing for a delta in floating point comparisons

pgrt:
I got failing tests on my machine and thus had to allow for a tiny delta between float of double numbers.

If I get it well, the comparison in CompatibleFieldSerializerTest.java is between long numbers, so another fix could be to use
        assertEquals(1L, o.value1);
The assertEquals method with 2 arguments can be applied to long numbers. With 3 arguments it is only for float or double (so in the current code, there is an implicit conversion from long to a floating-point type).

Best,
Pierre

### DIFF
--- a/test/com/esotericsoftware/kryo/KryoAssert.java
+++ b/test/com/esotericsoftware/kryo/KryoAssert.java
@@ -26,15 +26,18 @@ import static org.junit.jupiter.api.Assertions.*;
  */
 public class KryoAssert {
 
+    static final double EPS_D = 1.E-14;
+    static final float EPS_F = 5.E-7f;
+
     private KryoAssert() {
 
     }
 
     public static void assertDoubleEquals(double expected, double actual) {
-        assertEquals(expected, actual, 0.0);
+        assertEquals(expected, actual, EPS_D);
     }
 
     public static void assertFloatEquals(float expected, double actual) {
-        assertEquals(expected, actual, 0.0);
+        assertEquals(expected, actual, EPS_F);
     }
 }
--- a/test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
+++ b/test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
@@ -43,6 +43,8 @@ class CompatibleFieldSerializerTest extends KryoTestCase {
 		supportsCopy = true;
 	}
 
+        static final double EPS = 1.E-14;
+
 	@Test
 	void testCompatibleFieldSerializer () {
 		testCompatibleFieldSerializer(83, false, false);
@@ -264,7 +266,7 @@ class CompatibleFieldSerializerTest extends KryoTestCase {
 		otherKryo.register(ClassWithWrapperAndPrimitive.class, otherSerializer);
 
 		ClassWithWrapperAndPrimitive o = (ClassWithWrapperAndPrimitive) otherKryo.readClassAndObject(input);
-		assertEquals(1L, o.value1, 0);
+		assertEquals(1L, o.value1, EPS);
 		assertEquals(1, o.value2);
 	}
 
--- a/test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
+++ b/test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
@@ -43,8 +43,6 @@ class CompatibleFieldSerializerTest extends KryoTestCase {
 		supportsCopy = true;
 	}
 
-        static final double EPS = 1.E-14;
-
 	@Test
 	void testCompatibleFieldSerializer () {
 		testCompatibleFieldSerializer(83, false, false);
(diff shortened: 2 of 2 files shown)


test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
```diff
@@ -264,7 +266,7 @@ private void testChangePrimitiveAndWrapperFieldTypes (int length, boolean chunke
 		otherKryo.register(ClassWithWrapperAndPrimitive.class, otherSerializer);
 
 		ClassWithWrapperAndPrimitive o = (ClassWithWrapperAndPrimitive) otherKryo.readClassAndObject(input);
-		assertEquals(1L, o.value1, 0);
+		assertEquals(1L, o.value1, EPS);
```
theigl, line 269: Please remove the delta argument as you suggested.

test/com/esotericsoftware/kryo/KryoAssert.java
```diff
@@ -26,15 +26,18 @@
...
 
     public static void assertDoubleEquals(double expected, double actual) {
-        assertEquals(expected, actual, 0.0);
+        assertEquals(expected, actual, EPS_D);
     }
 
     public static void assertFloatEquals(float expected, double actual) {
-        assertEquals(expected, actual, 0.0);
+        assertEquals(expected, actual, EPS_F);
     }
```
theigl, line 42: Do all of the double and float asserts fail in your case, or only a fraction of them? If only some asserts are affected we could add the delta only where it is needed.

theigl (changes requested):

test/com/esotericsoftware/kryo/serializers/CompatibleFieldSerializerTest.java
pgrt: This is now done in my second commit, I used the assertEquals(long, long) form and the longValue() method to get a long from a Long.

test/com/esotericsoftware/kryo/KryoAssert.java
pgrt: I got many failures but did not check if all such assertions were failing, or only some of them. Would you be interested in the complete list of the failures I got? If so, I can remove my patch and run the tests again.

theigl: One example each for a failing float and a failing double comparison would be great.

Just an idea: Would removing the delta from these two methods also resolve your problem?

pgrt: Without my patch I got

[ERROR] Failures: 
[ERROR]   InputOutputTest.testDoubles:742->runDoubleTest:791 positive delta expected but was: <0.0>
[ERROR]   InputOutputTest.testFloats:649->runFloatTest:697 positive delta expected but was: <0.0>
[ERROR]   UnsafeByteBufferInputOutputTest.testDoubles:700->runDoubleTest:750 positive delta expected but was: <0.0>
[ERROR]   UnsafeByteBufferInputOutputTest.testFloats:606->runFloatTest:655 positive delta expected but was: <0.0>
[ERROR]   UnsafeInputOutputTest.testDoubles:666->runDoubleTest:716 positive delta expected but was: <0.0>
[ERROR]   UnsafeInputOutputTest.testFloats:572->runFloatTest:621 positive delta expected but was: <0.0>
[ERROR]   CompatibleFieldSerializerTest.testChangePrimitiveAndWrapperFieldTypes:248->testChangePrimitiveAndWrapperFieldTypes:267 positive delta expected but was: <0.0>

There are 3 failures related to flats, 3 related to doubles, and the last one is the one concerning longs.

About your suggestion: I see in the doc of JUnit4 that assertEquals(double, double) has been deprecated and that one should ``Use assertEquals(double expected, double actual, double delta) instead''. So maybe we should stick to the positive delta fix.

theigl:
The change looks good now, but I'm a bit reluctant to merge it because the delta might hide potential future encoding issues.

I would like to know what causes the issues on your system. Maybe we can detect it and only enable the delta for this specific scenario? You said you are using Debian, but I'm not sure if this really is the cause. What kind of architecture is your CPU and which JVM vendor do you use?

pgrt:
I understand your point. I think merging with the smallest possible delta makes sense (1.E-14 could be lowered), but I understand you want to dig more.

I got the issue in a minimal Debian system with an AMD64 cpu, with openjdk-11. Would you see any other information I could provide?
Hope this helps...
Pierre

theigl:
@pgrt: I just tried to replicate your issue on a fresh Debian unstable install.

This is what I did:

1. `docker pull debian:unstable`
2. `docker run -it $CONTAINER /bin/sh`
3. `apt update`
4. `apt install git openjdk-11-jdk maven`
5. `git clone https://github.com/EsotericSoftware/kryo.git`
6. `mvn clean && mvn install`

The build succeeded without any issues. Is your setup any different?

pgrt:
Hi,

I got the failures in a chroot of Debian unstable created using pbuilder. Then I ran debhelper scripts to build Debian packages, you can find enclosed the log with relevant parts and more... for instance the invocation of the tests is done at line 2473, and the errors are then below.
I hope this can help, although I am unsure this will provide a solution :-(

If you cannot reproduce the issue and if no one else complains about it, then maybe you can close my PR and just keep this potential issue in mind in case it happens again in the future.

Best regards,
Pierre
[log.txt](https://github.com/EsotericSoftware/kryo/files/6575933/log.txt)

theigl:
Thanks for your feedback @pgrt!

> If you cannot reproduce the issue and if no one else complains about it, then maybe you can close my PR and just keep this potential issue in mind in case it happens again in the future.

I'll close the issue and PR for now and will re-open it if someone else runs into this issue or can provide more info why/when this is happening.

PR #835, opened 2021-06-14
optimize unnecessary require operation in write_ascii_slow

pyb1993:
you can see the [discussion](https://groups.google.com/g/kryo-users/c/TjZik6HfQls/m/dESWdK43AgAJ) here
This PR is to reduce the require operation when serialize huge ascii string
You can use  

> mvn -f benchmarks/pom.xml compile exec:java -Dexec.args="-f 4 -wi 5 -i 3 -t 2 -w 2s -r 2 HugeStringBenchmark.writeAsciiHuge"
to test the performance (with and without this optimization), I have show a short result in the discussion

### DIFF
--- a/benchmarks/src/main/java/com/esotericsoftware/kryo/benchmarks/io/HugeStringBenchmark.java
+++ b/benchmarks/src/main/java/com/esotericsoftware/kryo/benchmarks/io/HugeStringBenchmark.java
@@ -1,0 +1,52 @@
+/* Copyright (c) 2008-2020, Nathan Sweet
+ * All rights reserved.
+ * 
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
+ * conditions are met:
+ * 
+ * - Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the distribution.
+ * - Neither the name of Esoteric Software nor the names of its contributors may be used to endorse or promote products derived
+ * from this software without specific prior written permission.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+ * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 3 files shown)


theigl:
@pyb1993: I have not forgotten about this PR. I'm currently on vacation and will take a look when I'm back in 2 weeks.

theigl:
@pyb1993: I just looked into your PR and can reproduce the performance gains. However, I'm not sure about applying your changes, since they only affect a rare corner case. As you wrote on the mailing list:

> But If there is such a case(and I do a benchmark for this rare case), the performance will have huge improvement in writeAscii.
> 1. the length of string is larger than the initial output buffer size. (which will cause the unecessary require operation(allocate and copy))
> 2. the output is not reused, new output each time (I think the best practice is reused the output if possible)

If performance is important, the output should be pooled and re-used. Otherwise you will see a lot of allocation and a lot of GC pressure. Your change is simple enough, but it still makes the code slightly harder to understand and I'm not sure if any real user would benefit from the it.

pyb1993:
Hi，I know this is really a rare case，but I found some other case that this function “maxallowedRequired” may be used，such as the writeInts method，which directly use capcacity to compare a threshold。


maybe there are more exra case that can benefit from this function。this function
should be a general function to be used in such situation。

------------------ Original ------------------
From: Thomas Heigl ***@***.***&gt;
Date: Mon,Jul 5,2021 7:10 PM
To: EsotericSoftware/kryo ***@***.***&gt;
Cc: .... ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [EsotericSoftware/kryo] optimize unnecessary require operation in write_ascii_slow (#835)





 
@pyb1993: I just looked into your PR and can reproduce the performance gains. However, I'm not sure about applying your changes, since they only affect a rare corner case. As you wrote on the mailing list:
  
But If there is such a case(and I do a benchmark for this rare case), the performance will have huge improvement in writeAscii.
  
the length of string is larger than the initial output buffer size. (which will cause the unecessary require operation(allocate and copy)) 

the output is not reused, new output each time (I think the best practice is reused the output if possible) 
  
If performance is important, the output should be pooled and re-used. Otherwise you will see a lot of allocation and a lot of GC pressure. Your change is simple enough, but it still makes the code slightly harder to understand and I'm not sure if any real user would benefit from the it.

pyb1993:
> @pyb1993: I just looked into your PR and can reproduce the performance gains. However, I'm not sure about applying your changes, since they only affect a rare corner case. As you wrote on the mailing list:
> 
> > But If there is such a case(and I do a benchmark for this rare case), the performance will have huge improvement in writeAscii.
> > 
> > 1. the length of string is larger than the initial output buffer size. (which will cause the unecessary require operation(allocate and copy))
> > 2. the output is not reused, new output each time (I think the best practice is reused the output if possible)
> 
> If performance is important, the output should be pooled and re-used. Otherwise you will see a lot of allocation and a lot of GC pressure. Your change is simple enough, but it still makes the code slightly harder to understand and I'm not sure if any real user would benefit from the it.

Hi, how do you think about his PR? as what I talked above, I think there are a lot similar situation that 
`capacity >= count << k` in writeLongs, writeInts, and writeDoubles

theigl:
@pyb1993: I'm still +0 on this. It can improve things for edge cases, but I'm not sure the change is worth it. I'll try to get some more feedback on this.

PR #839, opened 2021-06-30
avoid flush repeatedly when has finished flushing

jackjoesh:
Modify content： avoid flush repeatedly when has finished flushing.  
After my testing ,this modification can save cpu usage in high concurrency service.

### DIFF
--- a/src/com/esotericsoftware/kryo/io/OutputChunked.java
+++ b/src/com/esotericsoftware/kryo/io/OutputChunked.java
@@ -1,15 +1,15 @@
 /* Copyright (c) 2008-2020, Nathan Sweet
  * All rights reserved.
- *
+ * 
  * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
  * conditions are met:
- *
+ * 
  * - Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


src/com/esotericsoftware/kryo/io/OutputChunked.java
```diff
@@ -58,8 +58,9 @@ public void flush () throws KryoException {
 			} catch (IOException ex) {
 				throw new KryoException(ex);
 			}
+		}else{
```
theigl, line 61: Please apply project formatting to your changes. There should be whitespace around the `else`.

theigl (changes requested):
Apart from minor formatting, this PR looks good to me.

theigl:
Thanks @jackjoesh! I think this change makes sense and I will apply it.

src/com/esotericsoftware/kryo/io/OutputChunked.java
jackjoesh: thank you , i have formtted

jackjoesh:
> Thanks @jackjoesh! I think this change makes sense and I will apply it.

thank you for checking

theigl:
Thanks again @jackjoesh!

Issue #864, opened 2021-11-07
ArrayIndexOutOfBoundsException when serializing generic lambda written in Kotlin


holgerbrandl:
**Describe the bug**
When serializing an object in Kotlin it fails with an ArrayIndexOutOfBoundsException
```
Exception in thread "main" com.esotericsoftware.kryo.KryoException: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
Serialization trace:
definition (org.koin.core.definition.BeanDefinition)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:101)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.write(FieldSerializer.java:108)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:711)
	at KryoBeanDefinitionKt.main(KryoBeanDefinition.kt:31)
	at KryoBeanDefinitionKt.main(KryoBeanDefinition.kt)
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at com.esotericsoftware.kryo.util.DefaultGenerics.pushTypeVariables(DefaultGenerics.java:116)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.pushTypeVariables(FieldSerializer.java:147)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.write(FieldSerializer.java:102)
	at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:642)
	at com.esotericsoftware.kryo.serializers.ReflectField.write(ReflectField.java:70)
	... 4 more

Process finished with exit code 1
```

**To Reproduce**
I've created a minimal example in https://github.com/holgerbrandl/kryo-koin. The respective class to reproduce the stacktrace from above is `KryoBeanDefinitionKt#main`

**Environment:**
 - OS: Windows 10
 - JDK Version: 11
 - Kryo Version: 5.2.0

**Analysis**
It seems to be a mismatch between the array sizes of `args` and `hierarchy` in  
https://github.com/EsotericSoftware/kryo/blob/76c5e5d967a8d4a7e438ffcc4dd0be825ca93ce3/src/com/esotericsoftware/kryo/util/DefaultGenerics.java#L116
![image](https://user-images.githubusercontent.com/200952/140645833-541769ad-9378-4aa5-b0e2-6f07b5e1b5b5.png)

[...]

PR #879, opened 2022-01-26
[DRAFT] Add `MinimalGenerics` and allow users to set a `GenericsStrategy`

theigl:
We currently support wo strategies for generics: `DefaultGenerics` and `NoGenerics`. 

The former is enabled by default, but was responsible for most of the bugs found in Kryo 5 and incurs some performance overhead. Users can currently opt-out of generics optimization by setting `Kryo.setOptimizeGenerics(false)`. Internally, this flag sets the generic strategy to `NoGenerics`.

This PR introduces a third strategy called `MinimalGenerics` that only processes generic type information, but does not attempt to resolve type variables. This makes sense because the type variable logic is quite involved, still has edge-case issues (e.g. #864), and causes most of the performance overhead related to generics handling.

The new strategy is safer and should theoretically be faster than disabling generics optimizations altogether, because concrete generic types can still be resolved. `CollectionSerializer` and `MapSerializer` should especially benefit from this.

TODO

- [ ] JavaDoc for strategies and configuration option
- [ ] Find a better name for new generics strategy


NathanSweet:
Ideally full generics would be efficient and not have bugs for edge cases (at least #864 is resolved). If the performance overhead can't be improved then I'm fine having a `MinimalGenerics` implementation that is roughly what we had in Kryo 4. Though IIRC the performance difference from Kryo 4 to 5 was only marginally in 4's favor. Note that using generics information can remove the need to write types in many cases, so the trade with minimal/no generics is less CPU for larger serialized size.

NathanSweet:
BTW I got here from the [V6 ideas](https://github.com/EsotericSoftware/kryo/wiki/Kryo-v6-Ideas) page. All the other ideas sound good!

PR #886, opened 2022-02-21
Improve tests for JDK 17

theigl:
This PR attempts to fix as many tests as possible when running under JDK17 without any `add-opens` arguments.

Most tests work without opening up modules now. Only one tests still fails:

- `SerializationCompatTest` fails because of `AtomicInteger`. 

We need custom serializers for `Atomic*` JDK classes to fix this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -618,12 +618,11 @@ Kryo `isFinal` is used to determine if a class is final. This method can be over
 
 Kryo can serialize Java 8+ closures that implement java.io.Serializable, with some caveats. Closures serialized on one JVM may fail to be deserialized on a different JVM.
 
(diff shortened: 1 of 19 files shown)


theigl:
I added a fall-back for lambda deserialization in [f3cd2df](https://github.com/EsotericSoftware/kryo/pull/886/commits/f3cd2dfd7cbaff81f4c734d42a491f12555c5953).

theigl:
The only remaining issue is with `AtomicInteger` and we can't fix this without breaking backwards compatibility. We should add serializers for `Atomic*`, `UUID`, `Pattern` etc in Kryo 6.

Issue #884, opened 2022-02-14
VersionedRecordSerializer?


JanecekPetr:
**Is your feature request related to a problem? Please describe.**
I'm trying to serialize a Record that evolves - new components are being added to it.

**Describe the solution you'd like**
Would it make sense to create a `VersionedRecordSerializer` akin to `VersionedFieldSerializer` with the same functionality, the `@Since` annotation?

**Describe alternatives you've considered**
An obvious workaround is to use a traditional class instead of a record.

theigl:
@JanecekPetr: The `RecordSerializer` was a contribution from Oracle. When I accepted it, I didn't really think about aligning it with `FieldSerializer` and its compatible variants, but it definitely makes sense to be able to evolve records in a similar way to classes.

I don't have to time work on this at the moment but I'd be happy to accept a PR.

JanecekPetr:
Okay, I actually started looking into this.

1. The eclipse project setup is outdated - JUnit 5 was introduced, but the setup doesn't know about it yet. I'm not yet sure I want to provide a fix for that too, I'll see.
2. The current `RecordSerializer` is *sloooooow* as it does a lot of reflection all the time and no component caching. I started with a caching version (no ASM yet) and the speedup of writing is 20-25× (no read/roundtrip benchmarks yet, I'll do those next week):
    ```
    Benchmark                                               Mode  Cnt     Score    Error  Units
    RecordSerializerBenchmark.cachingRecordSerializer       avgt    5    85.466 ±  5.142  ns/op
    RecordSerializerBenchmark.recordSerializer              avgt    5  2035.110 ± 42.174  ns/op
    ```
    This is annoying as it's largely unfixable without breaking changes as there's now only `new RecordSerializer()` constructor and no `new RecordSerializer(Class<T> type)` constructor. What should I do? The new `VersionRecordSerializer` will for sure cache things and I'd be happy to fix the `RecordSerializer`, too, but how? What do you think is the best approach to the design now?
3. There is almost no configuration of the `RecordSerializer`, there's only `fixedFieldTypes`. I haven't yet looked into what other configuration might be necessary / useful as I do not yet fully understand all of them. Would you mind taking a look and helping me with the choice? If there is more config that'd be useful, I'll add that too. That, at least, can easily be done retroactively.

theigl:
@JanecekPetr: Thanks again for your valuable input!

[...]

PR #887, opened 2022-02-24
[884] Fix eclipse project setup: Add JUnit 5, ByteBuddy

JanecekPetr:
As discussed in #884:
- switched JUnit 4 for JUnit 5
- added ByteBuddy for the benchmarks project
- and, most controversially, switched to JDK 11 as there are classes using Java 9+ features. I can roll this back, but it makes a fair amount of classes uncompilable.


theigl:
@JanecekPetr:

> switched JUnit 4 for JUnit 5

JUnit 5 provides a single (unfortunately named) artifact as well:

> An executable junit-platform-console-{platform-version}-standalone.jar artifact containing the Jupiter and Vintage test engines with all dependencies is generated by the default build process, stored in junit-platform-console/build/libs and published to the Maven Central [repository](https://repo1.maven.org/maven2/org/junit/platform/junit-platform-console). It provides hassle-free usage of JUnit 5 in projects that manually manage their dependencies similar to the [plain-old JAR](https://github.com/junit-team/junit4/wiki/Download-and-Install#plain-old-jar) known from JUnit 4.

Could you try if this artifact works for the build? A single JAR would be much easier to manage.

> and, most controversially, switched to JDK 11 as there are classes using Java 9+ features. I can roll this back, but it makes a fair amount of classes uncompilable.

We should be safe, since our CI builds still run the tests on JDK 8.

JanecekPetr:
Fair enough, fixed!

theigl:
> We should be safe, since our CI builds still run the tests on JDK 8.

Sorry I wasn't really clear here: I meant, that you can change the JDK to 11 if you want, because our CI build will catch any incompatibilities with JDK 8 in the main source folder.

LGTM!

Issue #889, opened 2022-03-16
java.sql.Timestamp not persisted with full precision


purpledrazi:
**Describe the bug**
java.sql.Timestamp supports nanosecond precision, but data is truncated to milliseconds after being serialised and deserialised.

**To Reproduce**
```java
import static org.assertj.core.api.Assertions.assertThat;

import java.io.ByteArrayInputStream;
import java.io.ByteArrayOutputStream;
import java.sql.Timestamp;
import java.time.Instant;

import org.junit.jupiter.api.Test;

import com.esotericsoftware.kryo.Kryo;
import com.esotericsoftware.kryo.io.Input;
import com.esotericsoftware.kryo.io.Output;

class KryoTimestampTest {

  @Test
  void timestampRoundtrip() {
    
    var k = new Kryo();
    k.setRegistrationRequired(false);
    
    var bytes = new ByteArrayOutputStream();
    
    var originalTimestamp  = new Timestamp(Instant.parse("2022-03-16T17:03:05Z").toEpochMilli());
    originalTimestamp.setNanos(123456789);
    
    assertThat(originalTimestamp.toInstant()).isEqualTo(Instant.parse("2022-03-16T17:03:05.123456789Z"));
    
    try(var out = new Output(bytes)) {
      k.writeClassAndObject(out, originalTimestamp);
    }
    
    try(var in = new Input(new ByteArrayInputStream(bytes.toByteArray()))) {
      var deserialisedTimestamp = (Timestamp) k.readClassAndObject(in);
      assertThat(deserialisedTimestamp.toInstant()).isEqualTo(Instant.parse("2022-03-16T17:03:05.123456789Z"));
    }
  }
}
```

**Environment:**
 - OS: Debian
 - JDK Version: 17.0.2
 - Kryo Version: 5.3.0

**Additional context**
None

purpledrazi:
Raise PR #890 for this

PR #890, opened 2022-03-19
#889 Serializer for java.sql.Timestamp which preserves nanoseconds

purpledrazi:
I've added the new serializer to DefaultSeriailzers (even though it won't be used by default in Kryo 5). Is this OK? If not, let me know where it belongs.


theigl:
Please apply `mvn formatter:format`. Otherwise your PR looks good!

purpledrazi:
Updated with formatting applied.

theigl:
Merged, thank you @purpledrazi!

Issue #894, opened 2022-04-25
Will TaggedFieldSerializer be able to skip null fields?


xuedonglxd:
**Is your feature request related to a problem? Please describe.**
We have some POJO classes with hundreds of fields defined, but in each instance maybe there are only a few fields which are not null. (well, things are like this for a reason, different users could have different views of the same entity) 
And we used "chunked " just to be safe, because we ofter add new fields.
What I'm trying to say is, serializing (and then deserializing) those null fields could be a big waste.

**Describe the solution you'd like**
When writing fields, firstly iterate all the fields and record all the tags for non-null fields.
And only deal will non-null fields after that (so the number of fields will differ between different instances of the same class)
I have implemented this by my own by in a quite ugly way because it is almost impossible to inherit from the TaggedFieldSerializer class (because of its private fields).

So will you consider this or should I initiate a pull request of my version?

theigl:
@xuedonglxd: This sounds like a very specific use case. You can create a PR, but I cannot promise that I will merge it, if it adds too much complexity. What private fields do you need to access to implement this in your custom solution? Maybe we can expose them.

NathanSweet:
I'm not sure if others would make use of this specific feature. It might be slightly more interesting to provide a way of skipping any field with user defined criteria (eg by overriding a method) rather that only fields that are null. I'm not convinced it needs to go in TaggedFieldSerializer though. If `writeTags` was protected, a subclass could do it. However, it's not really all that terrible to just copy/paste TaggedFieldSerializer and change it for a particular use case.

PR #896, opened 2022-04-29
#894 Enable TaggedFieldSerializer to skip null fields (only works wit…

eridani40:
In my scenario, we have some really big classes with nearly a thousand fields, but each time we would probably just read 20~30 fields, based on our purposes. 
So serializing and deserializing the whole entity with chunked encoding could be a serious waste from the performance perspective. 

(Well, things are like this because those entities are really basic in our business, e.g video entity for Youtube; and their fields are managed by metadata systems, so they grow big really fast. But different fields are used by different businesses, and most like only a few fields at a time.)

I have done some tests, both the time consuming and the packet size are both considerably smaller then not skipping null fields. (and actually even for entities which are not so big ,the effect is quite obvious as well ,as long as there are null fields). I could share those test results if necessary. And I am not sure if I should add test cases as well for the new feature.

My solution is to filter the cached fields when trying to write the objects, and to do absolutely nothing when reading them.

I am not quite sure if it's a common case,  if it's done in a proper way, or in a correct form (it's my very first PR on github).
So just please take a look. Thanks.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/TaggedFieldSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/TaggedFieldSerializer.java
@@ -38,6 +38,9 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /** Serializes objects using direct field assignment for fields that have a <code>@Tag(int)</code> annotation, providing backward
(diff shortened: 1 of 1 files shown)


theigl:
@xuedonglxd: Thanks for your PR! 

The PR itself is good, but as I said in the ticket, I'm not sure this is a widely useful feature. 

It would be possible to introduce a hook like `protected getWriteTags(object)` that can be used to filter the fields before writing them. Then you would only have to override a single method to get your custom behavior. I'm a bit reluctant to change to contract of `TaggedFieldSerializer` just for this use case.

For now, let's keep this PR open and see if anyone else is interested in this.

theigl:
Closing because there has been no further interest in this PR.

PR #903, opened 2022-07-19
fix: input constructor call error

sunchaser-lilu:
Input class does not have a constructor with two int parameters


theigl:
Merged, thanks @sunchaser-lilu

PR #908, opened 2022-08-16
Return false if the buffer is not resized after flush.

fangzheng:
According to the doc string, `Output.require()` should return true if the buffer is resized and false otherwise. However, current implementation returns true if the buffer is not resized after `flush()`.

### DIFF
--- a/src/com/esotericsoftware/kryo/io/ByteBufferOutput.java
+++ b/src/com/esotericsoftware/kryo/io/ByteBufferOutput.java
@@ -182,7 +182,7 @@ public class ByteBufferOutput extends Output {
 	protected boolean require (int required) throws KryoException {
 		if (capacity - position >= required) return false;
 		flush();
-		if (capacity - position >= required) return true;
+		if (capacity - position >= required) return false;
 		if (required > maxCapacity - position) {
 			if (required > maxCapacity)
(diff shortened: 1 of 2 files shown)


theigl:
@fangzheng: Did you run into any issues with the current behavior, or do you simply want to fix the inconsistency in the docs?

I took a look at the [commit](https://github.com/EsotericSoftware/kryo/commit/df3e4dbac1a45aea008524cde34c1650f34c702b) introducing these lines. It seems that the method returns `true` for backwards-compatibility with the old code.

I'm a bit reluctant to change the implementation just for the sake of the docs.

fangzheng:
> @fangzheng: Did you run into any issues with the current behavior, or do you simply want to fix the inconsistency in the docs?
> 
> I took a look at the [commit](https://github.com/EsotericSoftware/kryo/commit/df3e4dbac1a45aea008524cde34c1650f34c702b) introducing these lines. It seems that the method returns `true` for backwards-compatibility with the old code.
> 
> I'm a bit reluctant to change the implementation just for the sake of the docs.

HI @theigl , i did not run into any issue with it, just feel it's inconsistent by reading the code. If it's for backwards compatibility, i am ok with it. Thanks.

fangzheng:
Maybe a comment to note backwards compatibility would be helpful, especially for anyone who wants to implement a derived class.

Issue #912, opened 2022-09-07
optimize serialization for primitive arrays [draft]


jhsenjaliya:
**Is your feature request related to a problem? Please describe.**

primitive arrays pre-occupies space as per the size of the elements, but its perfectly possible that most of the space is not used while serializing

**Describe the solution you'd like**
Instead of writing `size` and data from 0 to `size`, 
Improvement proposal is to serialize array as per following

`size` -- defined size for the array to be serialized, as it is today
if(primitive_array_optimization_configured){
    `actual_size` -- this is the actual size of array where user has set any value.
    `data` -- from 0 to `actual_size` 
}else{
    `data` -- from 0 to `size` 
}

where `actual_size` is calculated by reducing size until non default value is detected ( ex: 0 for int array )

**Describe alternatives you've considered**
None available to extend the array serializer

**Additional context**
Having this feature can greatly benefit size of the serialized object where primitive arrays are being used. 

Happy to hear comments and suggestions.

theigl:
The disadvantage is that you will have to traverse the array multiple times, or write the data in reverse order. I'm not sure this makes sense as a general addition to Kryo, but you should have no problem writing such custom `SparseArraySerializers` yourself and using them instead of the default implementation.

jhsenjaliya:
actually u would only need to traverse the array once ( from end until u find non-default value), and this can be activated with configuration so default behavior wont change.

theigl:
Please create a PR with an implementation for one of the [default array serializers](https://github.com/EsotericSoftware/kryo/blob/cb255af4f8df4f539778a325b8b4836d41f84de9/src/com/esotericsoftware/kryo/serializers/DefaultArraySerializers.java#L63). I'm still not sure this should be provided by Kryo out of the box, but we can discuss it further.

theigl:
Closing because there has been no interest in this so far.

PR #915, opened 2022-09-15
[issues-912] optimize serialization size for primitive arrays - IntOnly for now

jhsenjaliya:
PR for #912 ( Currently limited to showcase the feature for Int primitive arrays only based on the guidelines )

Motivation:
Many times, primitive array is used for optimized/efficient code (size and execution time), where the size of the primitive array is not always accurately decided beforehand. 
Serializing these primitive arrays can produce large serialized object with mostly blank character (NUL).
Kryo being focused in efficiency, should provide configuration to further optimize this behavior to store only the necessary values whenever there is an opportunity to do so.

pls refer to the description on #912 for more info on approach.

Thanks

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -162,6 +162,7 @@ public class Kryo {
 	private IdentityMap originalToCopy;
 	private Object needsCopyReference;
 	private Generics generics = new DefaultGenerics(this);
+	public boolean optimizePrimitiveArrays = false;
 
 	/** Creates a new Kryo with a {@link DefaultClassResolver} and references disabled. */
 	public Kryo () {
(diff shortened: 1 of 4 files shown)


theigl:
@jhsenjaliya: Thanks for this PR. Your approach is interesting, but the changes required to support it for all types are quite invasive. I will keep this PR open for now to see if anyone else is interested in this optimization.

jhsenjaliya:
Sure, I will let you think through this, its only for primitive type. 
My observation has been that optimizations will always be tricky, but the value it provides to have least possible storage size would be not only worth but adds lot of value to Kryo. 
Thanks for the review !

NathanSweet:
I agree the feature can be useful. I've used skipping zeros at the beginning or end in my projects, where it makes sense.

I don't think we want a setting that changes the behavior of all Input/Output. The feature can be entirely self contained within a serializer. Where you want this, which is unlikely to be everywhere, you would use the serializer.

It could make sense for Kryo to provide such a serializer, though there are many use case specific serializers that could be provided. We don't try to provide them all, especially when the implementation is relatively trivial.

If you really wanted to do it everywhere, you could extend Input/Output, but I don't think it makes sense for Kryo to provide that as it's too application specific.

jhsenjaliya:
@NathanSweet , Thanks for providing that input.
I believe such optimizations better suit as settings/configs rather than all new serializer/deserializer all together. 
Also by default this config is OFF, so there is no change to the behavior, only when user needs, it can be turned ON, when user thinks S/He wants additional improvements like this.

I also like ur idea of doing this for continuous default values ( zeros ) in the beginning instead of just end.  may be there can be settings/config for both -- optimize_continuous_zeros_in_starting_only, optimize_continuous_zeros_in_end_only and optimize_continuous_zeros ( for both )

I can imaging lot of storage savings with this. hope more people finds this useful feature/optimizations when needed.

theigl:
Closing because there has been no interest in this PR.

PR #927, opened 2022-12-02
Cache components, getters and constructors in `RecordSerializer`

theigl:
This PR adjusts `RecordSerializer` to cache record components, component getters and the canonical constructor. Thew new implementation is about 20x faster than the current one.

The only downside is that I had to remove the existing no-args constructor. This shouldn't be a huge issue, because there was no need for any user to instantiate the `RecordSerializer` themselves.

| Benchmark                        |   Mode |   Cnt    |     Score |   Error  |  Units |
| ------------- | ------------- | ------------- | -------------: | -------------: | ------------- |
| RecordSerializerBenchmark.new   |    thrpt |    4  | 2215445,525 | ± 20885,922 |  ops/s | 
| RecordSerializerBenchmark.old   |    thrpt  |   4  |  105476,509 | ±    1699,820 |   ops/s | 

If we want to stay fully backwards compatible, we can use `ClassValue` for caching record components and constructors.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/RecordSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/RecordSerializer.java
@@ -68,13 +68,13 @@ public class RecordSerializer<T> extends ImmutableSerializer<T> {
 		GET_TYPE = getType;
 	}
 
-	private static final ClassValue<Constructor<?>> CONSTRUCTOR = new ClassValue<>() {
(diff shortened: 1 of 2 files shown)


theigl:
@JanecekPetr: One last ping. Do you have any feedback on this implementation? If not, I'm going to merge this next week and release Kryo 5.4.0.

PR #931, opened 2022-12-16
[POC] FieldSerializer for Records

theigl:
This is a POC for moving record serialization from `RecordSerializer` into `FieldSerializer`.

This approach has two main advantages:

1. Existing configuration options for the `FieldSerializer` can be used for records
2. Sub-classes of `FieldSerializer` (Compatible, Versioned, Tagged, ..) can be adjusted for records

JDK 17+ is required for this PR so it is a likely candidate for the next major version.


Michal-Orlowski-KH:
Any plans to merge that one? :) I'm also waiting for that PR merge

Issue #943, opened 2023-03-19
Allow turning a few optimisations off for Kryo 5.4.0


chrisr3:
We are upgrading our code-base from Kryo 4.0.2 to 5.4.0, and have hit an issue when serialising a `Collection` only containing instances of `UnmodifiableRandomAccessList`.

We have sub-classed `Kryo` to support serialising classes with `writeResolve` and `readReplace` methods; `UnmodifiableRandomAccessList` implements `writeResolve` like this:
```java
private Object writeReplace() {
    return new UnmodifiableList<>(list);
}
```
Kryo 5's `CollectionSerializer` seems to have an optimisation for the case when all of collection's elements are of the same type, whereby Kryo serializes the element type once, and then serializes all of the elements' object data afterwards. Unfortunately, `UnmodifiableRandomAccessList.writeResolve` transforms the elements into instances of `UnmodifiableList`, but Kryo is unable to discover this until is has invoked `writeResolve` at least once. This is _after_ it has incorrectly written `UnmodifiableRandomAccessList` as the element type to the output stream, of course.

I cannot see a way of working around this apart from ignoring `UnmodifiableRandomAccessList.writeResolve`, although I cannot ignore _every_ `writeResolve` method because that would break lambda serialization. Ideally, I'd like to be able to configure `CollectionsSerializer` _not_ to apply this optimisation at all.

theigl:
Hi @chrisr3. Could you create a minimal reproducer (ideally a unit test) for the issue you are facing?

I'm open to adding a flag to `CollectionSerializer`, but there might be other solutions.

chrisr3:
This sample project should demonstrate what is happening:
[kryo-example.tar.gz](https://github.com/EsotericSoftware/kryo/files/11026870/kryo-example.tar.gz)

theigl:
OK thanks! 

I quickly glanced over the code:

Do you really need to globally override Kryo's behavior? Or would it be enough to create a custom serializer that uses `writeResolve/readReplace` internally and register it only for the classes where you actually want this behavior?

Lambda serialization is handled by `ClosureSerializer` that uses `readResolve/writeReplace` internally. So lambda serialization won't break if you change your global serialization logic.

chrisr3:
I think the problem is that _any_ class (including unknown "third party" ones) can (potentially) implement `readReplace` or `writeResolve`, so we cannot possibly know them all in advance.

chrisr3:
[...]

PR #947, opened 2023-03-24
Add a hook to intercept registration of all new serializers.

chrisr3:
Create a mechanism for customising, configuring, or even replacing a serializer when it is registered.

See #943.

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -437,6 +437,13 @@ public class Kryo {
 		return lowest;
 	}
 
+	/** A single point of contact for customising new serializer instances when they are registered.
+	 * Invoked by {@link DefaultClassResolver#register(Registration)}.
+	 */
+	public Serializer adaptNewSerializer (Serializer serializer) {
+		return serializer;
+	}
+
 	/** Returns the best matching serializer for a class. This method can be overridden to implement custom logic to choose a
 	 * serializer. */
 	public Serializer getDefaultSerializer (Class type) {
@@ -527,7 +534,7 @@ public class Kryo {
 	 * <p>
 	 * IDs must be the same at deserialization as they were for serialization.
 	 * <p>
-	 * Registration can be suclassed to efficiently store per type information, accessible in serializers via
+	 * Registration can be subclassed to efficiently store per type information, accessible in serializers via
 	 * {@link Kryo#getRegistration(Class)}. */
 	public Registration register (Registration registration) {
 		int id = registration.getId();
--- a/src/com/esotericsoftware/kryo/util/DefaultClassResolver.java
+++ b/src/com/esotericsoftware/kryo/util/DefaultClassResolver.java
@@ -26,6 +26,7 @@ import com.esotericsoftware.kryo.ClassResolver;
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.KryoException;
 import com.esotericsoftware.kryo.Registration;
+import com.esotericsoftware.kryo.Serializer;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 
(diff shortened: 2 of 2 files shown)


theigl:
Hi @chrisr3. Do you need this change for your custom serializer registration to work, or is this just something that would be nice to have? Would overriding `register` be enough in your case?

chrisr3:
> Hi @chrisr3. Do you need this change for your custom serializer registration to work, or is this just something that would be nice to have? Would overriding `register` be enough in your case?

Something like this would simplify my integration. The problem with overriding `Kryo.register(Registration)` is that not everything uses it:
```java
public Registration register (Class type, Serializer serializer) {
	Registration registration = classResolver.getRegistration(type);
	if (registration != null) {
		registration.setSerializer(serializer);
		return registration;
	}
	return classResolver.register(new Registration(type, serializer, getNextRegistrationId()));
}
```
And not forgetting
```java
public Registration register (Class type)
```
which invokes `register(Class, Serializer)`.

theigl:
I mean something like this:

```java
public class MyClassResolver extends DefaultClassResolver {

    @Override
    public Registration register(Registration registration) {
        return super.register(adapt(registration));
    }

    private Registration adapt(Registration registration) {
        // modify registration here
        return registration;
    }   
}
```

WDYT?

chrisr3:
> WDYT?

I think my customised `Kryo` class must now be paired with a customised `ClassResolver`, and _Bad Things_ will happen if they are separated.

And you presumably mean?
```java
protected Registration adapt(Registration registration)
```

theigl:
Do you have full control over the Kryo instance? Or can it be changed by clients/users? If you have full control, then using a custom class resolver isn't really an issue. It's just like any other configurable part of Kryo.

chrisr3:
We are _already_ providing Kryo with a modified `ClassResolver`, for unrelated reasons and from a different code-base. Which is why I would prefer to keep all of my Kryo-related changes encapsulated within the Kryo class.

theigl:
@chrisr3:

I gave your issue more thought over the weekend:

Event if I were to merge this change, you would still need a lot of complex code to achieve your desired behavior. Including your `CollectionSerializerAdapter`. I'm reluctant to add any code that only a single user needs and that doesn't even elegantly solve their problem.

I see only one option that does not require you to write a lot of workarounds: Adding something like `Kryo.setOptimizedCollections(false)`. Both `CollectionSerializer` and `ObjectArraySerializer` could check this flag and disable the optimization.

Again, this is most likely not a feature a lot of users are going to need. It adds to Kryo's public API and will have to be maintained in the future. On the other hand, it would be straightforward to implement and could even be made `protected` so only users who extend Kryo can set the flag.

chrisr3:
> @chrisr3: 
> Even if I were to merge this change, you would still need a lot of complex code to achieve your desired behavior. Including your `CollectionSerializerAdapter`. I'm reluctant to add any code that only a single user needs and that doesn't even elegantly solve their problem.

To be fair, I never intended to add my `CollectionSerializerAdapter` to this PR. I was simply looking to create an "officially supported" method of intercepting the creation of new serializers so that I could adapt `CollectionSerializer`. (The `ObjectArraySerializer` has no sub-classes, and so I can simply override it with my own rather then adapting it.)

> I see only one option that does not require you to write a lot of workarounds: Adding something like `Kryo.setOptimizedCollections(false)`. Both `CollectionSerializer` and `ObjectArraySerializer` could check this flag and disable the optimization.
> 
> Again, this is most likely not a feature a lot of users are going to need. It adds to Kryo's public API and will have to be maintained in the future.

Yes, I agree. This is why I was only trying to create a more general extension API that my adapter could use.

theigl:
> To be fair, I never intended to add my CollectionSerializerAdapter to this PR. I was simply looking to create an "officially supported" method of intercepting the creation of new serializers so that I could adapt CollectionSerializer and its sub-classes. (The ObjectArraySerializer has no sub-classes, and so I can simply override it with my own rather then adapting it.)

I think we misunderstood each other slightly. ;) I understand that you don't plan to add your `CollectionSerializerAdapter` to this PR. My main concern was that if I merge this PR, you will still have to use a lot of complex and potentially brittle (reflection) code **on your side** to get this working. So I was wondering if we shouldn't go the other way and make the behavior of `CollectionSerializer` and `ObjectArraySerializer` configurable. 

But I just gave this a quick try and I'm not happy with the solution. Even if I add a configuration flag, `CollectionSerializer` still tries to find a concrete element serializer by [analyzing the collections generics](https://github.com/EsotericSoftware/kryo/blob/e7c37a982fe6bd33f42cac235df109a685e48f47/src/com/esotericsoftware/kryo/serializers/CollectionSerializer.java#L102). So if you don't turn off generics optimization, there is at least one other code path that potentially breaks your application. So I would need to add multiple conditions to the already complex serializer.

chrisr3:
> But I just gave this a quick try and I'm not happy with the solution. Even if I add a configuration flag, `CollectionSerializer` still tries to find a concrete element serializer by [analyzing the collections generics](https://github.com/EsotericSoftware/kryo/blob/e7c37a982fe6bd33f42cac235df109a685e48f47/src/com/esotericsoftware/kryo/serializers/CollectionSerializer.java#L102). So if you don't turn off generics optimization, there is at least one other code path that potentially breaks your application. So I would need to add multiple conditions to the already complex serializer.

We're already turning the generics optimisation off because we're using Kotlin - see #864. Actually, I think I should dig deeper into this issue... :thinking:.

theigl:
I'm closing this because there has been no interested in the PR in the meantime. Please create an issue if this is still relevant.

PR #950, opened 2023-04-04
Add a unit test for a Kotlin lambda class.

chrisr3:
This _appears_ to be enough to include a Kotlin JUnit case, although IntelliJ doesn't like it much :nauseated_face:.

### DIFF
--- a/pom-main.xml
+++ b/pom-main.xml
@@ -39,6 +39,41 @@
 				<artifactId>build-helper-maven-plugin</artifactId>
 			</plugin>
 			<plugin>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-maven-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>test-compile</id>
+						<goals>
+							<goal>test-compile</goal>
+						</goals>
+						<configuration>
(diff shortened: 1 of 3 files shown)


theigl:
I just pushed another solution to this: #949. 

Could you check if that works better for you in IntelliJ? For me it works quite well after running a Maven re-import.

chrisr3:
> Could you check if that works better for you in IntelliJ? For me it works quite well after running a Maven re-import.

IntelliJ seems to like this solution better than mine, although I'm still struggling to persuade it to execute the new test. However, I'm more accustomed to using IntelliJ with Gradle projects rather than Maven ones, so all my problems could _still_ be down to "user error"... :facepalm:.

theigl:
> IntelliJ seems to like this solution better than mine, although I'm still struggling to persuade it to execute the new test. However, I'm more accustomed to using IntelliJ with Gradle projects rather than Maven ones, so all my problems could still be down to "user error"... 🤦.

Thanks a lot for testing it! Running the test works for me, but I'll try once more with a newly imported project just to make sure.

theigl:
We went with a different implementation in https://github.com/EsotericSoftware/kryo/pull/949

PR #1000, opened 2023-07-27
add URISerializer.

ko-sasaki:
DefaultURISerializer was in the default serializer package, but was not set in Kryo.java, so it was added


theigl:
Unfortunately, we cannot do this in Kryo 5 because it would break backwards-compatibility for users who already have serialized URIs.

I will add it as a default serializer in [Kryo 6](https://github.com/EsotericSoftware/kryo/wiki/Kryo-v6-Ideas). Please register the serializer manually for now.

sasaki-kouhei:
So it was. Thank you and I look forward to the release of Kryo6.

PR #1015, opened 2023-10-14
Avoid allocation in `BigDecimal` serializer.

gdela:
This is a continuation of #1014, following the great advise from @NathanSweet to avoid allocation in the `BigDecimalSerializer` and use the `Long.numberOfLeadingZeros()` to be able to do so, the serialization/deserialization of `BigDecimal`s is now even faster (and the code got simpler):

avoid-allocation-option-A


gdela:
Seems this one will be superceeded by #1016 / #1018.

PR #1018, opened 2023-10-26
Avoid allocation in BigDecimal serializer with alternative way to write bytes [crlf corrected].

gdela:
This is a continuation of #1016 but with the same commits but squashed to one and with fixed the problem with CR/LF (line endings) in `DefaultSerializers`. For reasoning and performance comparison see the #1016.


NathanSweet:
Looks good! Cheers!

PR #1146, opened 2024-10-28
Fix #1145: Attempt Thread.currentThread().getContextClassLoader()

### LINKED ISSUE #1145: DefaultClassResolver: Attempt Thread.currentThread().getContextClassloader()
**Describe the bug**
We ran into this with Quarkus which uses 6 different classloaders.  We were able to fix it with...

```java
kryo.setClassLoader( Thread.currentThread().getContextClassLoader());
```

But shouldn't that be one of the classloaders tested by default?

```java
try {
    type = Class.forName(className, false, kryo.getClassLoader());
} catch (ClassNotFoundException ex) {
    // Fallback to current thread classloader
    try {
        type = Class.forName(className, false, Thread.currentThread().getContextClassLoader());
    } catch (ClassNotFoundException ex2) {
        // Fallback to Kryo's class loader.
        try {
            type = Class.forName(className, false, Kryo.class.getClassLoader());
        } catch (ClassNotFoundException ex2) {
            throw new KryoException("Unable to find class: " + className, ex);
        }
    }
}
```

melloware:
Fix #1145: Attempt Thread.currentThread().getContextClassLoader()

### DIFF
--- a/src/com/esotericsoftware/kryo/util/DefaultClassResolver.java
+++ b/src/com/esotericsoftware/kryo/util/DefaultClassResolver.java
@@ -182,8 +182,8 @@ public class DefaultClassResolver implements ClassResolver {
 						// Fallback to Kryo's class loader.
 						try {
 							type = Class.forName(className, false, Kryo.class.getClassLoader());
-						} catch (ClassNotFoundException ex2) {
-							throw new KryoException("Unable to find class: " + className, ex);
+						} catch (ClassNotFoundException ex3) {
+							throw new KryoException("Unable to find class: " + className, ex3);
 						}
 					}
 				}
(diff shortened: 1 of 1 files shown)


theigl:
I thought about this some more and I'm not sure if we should really go in this direction: 

This would fix your issue, but at a cost. Assuming that Kryo's default class loader can never load your classes, you will get an exception for every class that you load, every time you deserialize an object graph. And you wouldn't notice because everything works as expected. If your object graph is complex and contains a large number of classes, this could result in significant overhead because throwing exceptions is expensive.

melloware:
I don't disagree with that.  I didn't look what is kryo's default classloader and should it be `Thread.currentThread().getContextClassLoader()` ?

theigl:
The default classloader is defined in [Kryo.java](https://github.com/EsotericSoftware/kryo/blob/2ca7ab8749371d7250099ff46990aaf7d4f08ad2/src/com/esotericsoftware/kryo/Kryo.java#L145) as:

        private ClassLoader classLoader = getClass().getClassLoader();

This default works fine in most environments and you are the first user to raise a class loading issue in years. So I'd say that your use-case is a bit special and there already is a solution that you can use to configure Kryo accordingly.

Given my concerns mentioned above, I tend to keep things as they are now.

Issue #1022, opened 2023-11-02
General approach to Java collections? -> UnmodifiableCollections do not seem to be de/serializable using Kryo


RafaShobr:
Hi!
we are currently in the process of migrating from hessian to Kryo and so far its looking good, I really like Kryo :)
Our tests have shown, that not all java collection classes seem to be supported however.

For example trying to de/serialize unmodifiableCollection (Collection unmodifiableCollection = Collections.unmodifiableCollection(testList);) throws:
com.esotericsoftware.kryo.KryoException: Class cannot be created (missing no-arg constructor): java.util.Collections$UnmodifiableCollection

I wanted to ask if this is an oversight, or what the general guideline or approach of Kryo is regarding the standard collection classes of the JDK. I'm trying to find out if we can rely on the standard collections or if we need to be implementing custom serializers for some of those and if there would be an exhaustive list of in or excluded standard collection classes or something like that? :)

Thanks and regards,
Rafa

theigl:
@RafaShobr: Sorry for the slow reply. I was AFK for the whole month. 

Kryo does not have built-in support UnmodifiableCollections. You need a third party serializer. See #997.

Issue #997, opened 2023-07-21
SynchronizedMap can't de-serialized


mrniko:
**Describe the bug**
Instance of Collections.synchronizedMap can't be de-serialized properly. Here is the thrown exception.

```
java.lang.NullPointerException: Cannot enter synchronized block because "this.mutex" is null

	at com.esotericsoftware.kryo.serializers.MapSerializer.read(MapSerializer.java:236)
	at com.esotericsoftware.kryo.serializers.MapSerializer.read(MapSerializer.java:42)
	at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:880)
	at org.redisson.codec.Kryo5Codec$4.decode(Kryo5Codec.java:144)
```

**To Reproduce**
```java
        Map<String, String> m = new HashMap<>();
        m.put("1", "2");

        Kryo5Codec c = new Kryo5Codec();
        ByteBuf v = c.getValueEncoder().encode(Collections.synchronizedMap(m));
        Map<String, String> map = (Map<String, String>) c.getValueDecoder().decode(v, null);
```

**Environment:**
 - OS: [e.g. Ubuntu]
 - JDK Version: 17
 - Kryo Version: 5.5.0

**Additional context**

theigl:
Kryo doesn't have a built-in serializer for synchronized collections because it is impossible to write a safe serializer for them (i.e. a serializer that doesn't use reflection and doesn't rely on implementation details).

If you want to serialize synchronized and unmodifiable collections, you need to register the corresponding serializers from https://github.com/magro/kryo-serializers.

mrniko:
Can kryo convert them to a regular collections during serializtion?

theigl:
That is not possible, because the underlying collection could be *any* type of collection. It could be a `SortedMap` for instance. In that case you would lose your custom sort order when deserializing it.

Issue #913, opened 2022-09-12
deserialization of an iCal4j objects fails - missing no-arg constructor for immutable properties


overwerk:
Hi,

the deserialization of an net.fortuna.ical4j.model.Calendar (iCal4j lib) objects fails and I don't know how to handle this.
Any help is appreciated :-)

My kryo settings:

            kryo = new Kryo();
		kryo.setRegistrationRequired(false);
		kryo.setReferences(true);
		kryo.setDefaultSerializer(CompatibleFieldSerializer.class);
		kryo.setWarnUnregisteredClasses(true);
	        kryo.register(DateTime.class, new JodaDateTimeSerializer());
		kryo.register(java.util.UUID.class, new UUIDSerializer());
		kryo.register(net.fortuna.ical4j.model.component.VEvent.class, new JavaSerializer());

The exception:

		00:02  WARN: Class is not registered: net.fortuna.ical4j.model.property.CalScale$ImmutableCalScale
		Note: To register this class use: kryo.register(net.fortuna.ical4j.model.property.CalScale.ImmutableCalScale.class);
		com.esotericsoftware.kryo.KryoException: Class cannot be created (missing no-arg constructor): 												net.fortuna.ical4j.model.property.CalScale$ImmutableCalScale
		Serialization trace:
		properties (net.fortuna.ical4j.model.Calendar)
		calendar (core.Calendar)
		calendar (core.EstateManager)
			at com.esotericsoftware.kryo.util.DefaultInstantiatorStrategy.newInstantiatorOf(DefaultInstantiatorStrategy.java:111)
			at com.esotericsoftware.kryo.Kryo.newInstantiator(Kryo.java:1190)
			at com.esotericsoftware.kryo.Kryo.newInstance(Kryo.java:1199)
			at com.esotericsoftware.kryo.serializers.FieldSerializer.create(FieldSerializer.java:163)
			at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.read(CompatibleFieldSerializer.java:117)
			at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:877)
			at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:245)
			at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:44)
			at com.esotericsoftware.kryo.Kryo.readObject(Kryo.java:796)
[...]

Issue #909, opened 2022-08-16
Add Collections.unmodifiableList() to JdkImmutableListSerializer.


jvdvegt:
**Is your feature request related to a problem? Please describe.**
The list returned by `Collections.unmodifiableList(someList)` is not serializable when using the default Kryo serializers. 

**Describe the solution you'd like**
Default JDK classes should be serializable as much as possible without using external libraries.

**Describe alternatives you've considered**
When I create a copy of the class `ImmutableCollectionsSerializers.JdkImmutableListSerializer` and register it to be used for unmodifiable lists, they seem to work fine:
` kryo.register(Collections.unmodifiableList(new ArrayList<>()).getClass(), new CopyOfJdkImmutableListSerializer());`

Exposing instances of the `JdkImmutable*Serializer` classes so I can register them myself would be second-best, I suppose.

**Additional context**
The same holds for `Collections.unmodifiableSet` and probably for `Collections.unmodifiableMap`.

theigl:
The problem is that `Collections.unmodifiableList(someList)` may wrap an arbitrary list and it is impossible to determine the underlying list type without resorting to reflection. The same problem applies to sets and maps, but is further complicated by possible sorted implementations, where the comparator has to be serialized as well.

Your workaround only works for wrapped `ArrayList` types and as a side-effect changes the type of the list to `ImmutableCollections.ListN`.

Kryo only provides *safe* serializers, i.e. serializers that do not rely on implementation details and reflection. However, you can use the implementation provided by the kryo-serializers project: https://github.com/magro/kryo-serializers/blob/master/src/main/java/de/javakaffee/kryoserializers/UnmodifiableCollectionsSerializer.java

jvdvegt:
Thanks for the thorough explanation. I'm currently using the javakaffee serializer you mentioned, but as it causes warnings about reflective access I am trying to get rid of it.
I am not interested in the exact type of list wrapped by `unmodifiableList()`: deserializing into any unmodifiable/immutable list is fine. So the `JdkImmutableListSerializer` works for me, and it seems there are now two way to use it:
1. Write my own implementation and register it explicitly in Kryo, or
2. Use reflection to access the class in Kryo and register it explicitly in Kryo.

Would exposing an instance of the `JdkImmutableListSerializer` so I could register it explicitly be an option?

theigl:
Since most other serializers are currently `public` it shouldn't be a problem to expose them. If you want, you can create a PR for this.

[...]

PR #1154, opened 2024-12-08
Add serializers for unmodifiable and synchronized collections

theigl:
This PR adds serializers for JDK unmodifiable and synchronized collections.

This has been one of the most requested features for Kryo, but I have so far rejected all requests and redirected users to the [kryo-serializers](https://github.com/magro/kryo-serializers/tree/master/src/main/java/de/javakaffee/kryoserializers) project because it is not possible to implement these serializers without reflection. Since this comes up again and again an we already have the same issue with the `ClosureSerializer` that has been part of Kryo for many years, I decided to give up my opposition.

The serializers need to be registered manually for now to maintain backwards compatibility, but will be added as default serializers in Kryo 6.

Note: The unsafe logic will probably have to be replaced with Varhandles for JDK24+.

#689
#885
#909
#913
#997
#1022


nastra:
@theigl thanks for working on this. Do you already know which version might be shipped with this feature?

theigl:
Hi @nastra. I'm still unhappy about merging this because it uses terminally deprecated code in `Unsafe`, but I think I will include it in Kryo 5.7.0. The release will probably happen in June or July.

PR #1179, opened 2025-03-31
Add better feedback when using abstract classes in serialization

### LINKED ISSUE #857: Minor Improvement: Kryo.readObject should detect an abstract class and throw an IllegalArgumentException
This is a very minor improvement to provide more helpful error messages:

**Describe the solution you'd like**

Calls to Kryo.readObject(Input, Class) will throw an exception if passed an abstract class:

```
com.esotericsoftware.kryo.kryo5.KryoException: Error constructing instance of class: 
```

This of course makes sense but the error message isn't very helpful. It doesn't actually explain why the the class couldn't be constructed. The underlying java.lang.InstantiationException provides zero information.

A more helpful protocol might be for kryo.readObject would immediately detect that it's being asked to instantiate an abstract class and raise an IllegalArgumentException.

pvlov:
This should improve the feedback given to the user when supplying an abstract class to deserialise Objects.

Closes #857

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -765,6 +765,7 @@ public class Kryo {
 	public <T> T readObject (Input input, Class<T> type) {
 		if (input == null) throw new IllegalArgumentException("input cannot be null.");
 		if (type == null) throw new IllegalArgumentException("type cannot be null.");
+		if (Modifier.isAbstract(type.getModifiers())) throw new IllegalArgumentException("type cannot be an abstract class.");
 		beginObject();
 		try {
 			T object;
@@ -787,6 +788,7 @@ public class Kryo {
 		if (input == null) throw new IllegalArgumentException("input cannot be null.");
 		if (type == null) throw new IllegalArgumentException("type cannot be null.");
 		if (serializer == null) throw new IllegalArgumentException("serializer cannot be null.");
+		if (Modifier.isAbstract(type.getModifiers())) throw new IllegalArgumentException("type cannot be an abstract class.");
 		beginObject();
 		try {
 			T object;
@@ -809,6 +811,7 @@ public class Kryo {
 	public <T> T readObjectOrNull (Input input, Class<T> type) {
 		if (input == null) throw new IllegalArgumentException("input cannot be null.");
 		if (type == null) throw new IllegalArgumentException("type cannot be null.");
+		if (Modifier.isAbstract(type.getModifiers())) throw new IllegalArgumentException("type cannot be an abstract class.");
 		beginObject();
 		try {
 			T object;


pvlov:
I did not add the check for `Kryo::readClassAndObject(Input)` as I don't think the check makes sense there. I would add a test to make sure this is enforced, but there don't seem to be tests for Exceptions being thrown correctly apart from one or two in the io tests.

pvlov:
`EnumSet` seems to be abstract, yet it is fine to deserialise into that. I'm not sure how this should be dealt with.

theigl:
Hi @pvlov. Thanks for the PR!

We should take a different approach to solve this and keep the check out of the hot-path for serialization. The simplest solution would be to throw a more informative exception in [DefaultInstantiatorStrategy](https://github.com/EsotericSoftware/kryo/blob/master/src/com/esotericsoftware/kryo/util/DefaultInstantiatorStrategy.java). 

It is possible to do this without any overhead at all if we do the check only in the case where an exception is already thrown and replace it with a more informative one.

I suggest you start by creating a failing test-case and check where the exception is currently thrown.

pvlov:
Hi, thanks for the pointers! As you suggested I looked into the DefaultInstantiatorStrategy.

Here is what I found out during Testing:

When you try to roundtrip an abstract static member-class an `InstantiationException` is thrown in the code path that uses regular Reflection, which then gets turned into a `KryoException` with the error message seen in the issue. Similarly to the message given when using anonymous classes I also added two possible workarounds for the exception. I think these are sound, but would love to hear if this is what is recommended in this use-case.

When you try to roundtrip a non-member abstract class then you get an `InstantiatonError` along the lines of:

```java
java.lang.InstantiationError: com.esotericsoftware.kryo.serializers.TestAbstractClass
...stacktrace and so on
```
I also added the improved feedback here as the error stems from the same problem. 

While I don't think the tests add an enough value to keep them, I left them for the sake of making my thought process clearer when you review this PR. I can also clean the tests up if you think they are worth keeping.

test/com/esotericsoftware/kryo/serializers/TestAbstractClass.java
theigl, line 1: Please make this class a static inner class of your test class.

test/com/esotericsoftware/kryo/serializers/TestConcreteClass.java
theigl, line 1: Please make this class a static inner class of your test class.

test/com/esotericsoftware/kryo/serializers/SerializerErrorBehaviorTest.java
```diff
@@ -0,0 +1,49 @@
...
+    void testCustomObjectInsantiatorForAbstractClass() {
+        Registration registration = kryo.register(AbstractClass.class);
+        registration.setInstantiator(ConcreteClass::new);
+    }
+
+    @Test
+    void testNonMemberAbstractClassCannotBeSerialized() {
+        kryo.register(TestAbstractClass.class);
+        kryo.register(TestConcreteClass.class);
+        assertThrows(KryoException.class, () -> simpleRoundtrip(TestAbstractClass.class, new TestConcreteClass()));
```
theigl, line 33: Please assert on the error message as well, so we known that your changes actually work. Something like:

```java
KryoException thrown = assertThrows(KryoException.class, () -> simpleRoundtrip(TestAbstractClass.class, new TestConcreteClass()));
assertTrue(thrown.getMessage().contains("The type you are trying to serialize into is abstract");
```

src/com/esotericsoftware/kryo/util/DefaultInstantiatorStrategy.java
```diff
@@ -86,7 +92,13 @@ public Object newInstance () {
...
 						return constructor.newInstance();
 					} catch (Exception ex) {
-						throw new KryoException("Error constructing instance of class: " + className(type), ex);
+						StringBuilder message = new StringBuilder("Error constructing instance of class: " + className(type));
+						// Note: For Array and Primitive types the abstract bit is always set.
+						if (!type.isArray() && !type.isPrimitive() && Modifier.isAbstract(type.getModifiers())) {
+							message.append("\nNote: The type you are trying to serialize into is abstract. Kryo will not be able to create an instance of it. Possible solutions:\n");
+							message.append("You can either use a concrete subclass or use a custom ObjectInstantiator to create an instance.");
+						}
+						throw new KryoException(message.toString(), ex);
```
theigl, line 101: Please extract a method that creates the exception and re-use it above. Catching the `InstantiationError` is a clever way to avoid the check for abstract, but for maintainability, reducing duplication is more important.

test/com/esotericsoftware/kryo/serializers/SerializerErrorBehaviorTest.java
```diff
@@ -0,0 +1,49 @@
...
+import com.esotericsoftware.kryo.Registration;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import org.junit.jupiter.api.Test;
+
+import java.io.ByteArrayOutputStream;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class SerializerErrorBehaviorTest extends KryoTestCase {
```
theigl, line 14: You are testing instantiation more than serialization. Please move this test to `com.esotericsoftware.kryo` and rename it to something like `InstantiatiorTest`.

theigl (changes requested):
Thanks! I left some minor comments.

src/com/esotericsoftware/kryo/util/DefaultInstantiatorStrategy.java
theigl: Should we also check for `type.isInterface()`? Please add a test that tries to instantiate an interface to see what happens now.

pvlov: Yes, this error is also thrown when handling Interfaces, since [`every interface is implicitly abstract`](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html). I added a test for interfaces too.

test/com/esotericsoftware/kryo/serializers/SerializerErrorBehaviorTest.java
pvlov: Yeah true, I simplified the tests a lot to directly call the `DefaultInstantiatorStrategy` instead of using a `Kryo`.

pvlov:
About moving the classes into the Testclass as a member class: This would then lead to the code path using ReflectASM not being covered due to the check for non-static member classes. Putting it right into the same file as the Tests seems to be a good compromise, but I don't particularly like it. I also don't see any other good solution that would make sure we don't lose out on coverage.
```java 
boolean isNonStaticMemberClass = enclosingType != null && type.isMemberClass() && !Modifier.isStatic(type.getModifiers());
if (!isNonStaticMemberClass) ...
```
https://github.com/EsotericSoftware/kryo/blob/22a6495e6e1586485c212e61885a87d881c44e7b/src/com/esotericsoftware/kryo/util/DefaultInstantiatorStrategy.java#L53


The reason for the interface check being not at the same spot as for abstract classes is because any exception that occurs during the use of reflection is ignored. Trying to get the constructor of an interface will already fail there.
https://github.com/EsotericSoftware/kryo/blob/22a6495e6e1586485c212e61885a87d881c44e7b/src/com/esotericsoftware/kryo/util/DefaultInstantiatorStrategy.java#L93

theigl:
>  Putting it right into the same file as the Tests seems to be a good compromise, but I don't particularly like it. I also don't see any other good solution that would make sure we don't lose out on coverage.

That's fine with me. I wouldn't do that for production code, but for tests it is ok.

Could you please run `maven formatter:format` and commit the changes to your two files, if any? If unrelated files are getting formatted, simply discard the changes.

And your test class is missing the licence header.

Otherwise, LGTM.

pvlov:
Alright, done! Thanks for all the guidance in this PR, this was my first contribution to OSS so all the help made it quite enjoyable to contribute :).

theigl (approved):

pvlov:
Is there anything I need to do from my side? I can't merge this myself, so I assume this will be merged when the time is right? I'm not really familiar with how this usually works 😅

theigl:
> Alright, done! Thanks for all the guidance in this PR, this was my first contribution to OSS so all the help made it quite enjoyable to contribute :).

Thats great to hear. Thank you for choosing Kryo for your first contribution 😄 

> Is there anything I need to do from my side? I can't merge this myself, so I assume this will be merged when the time is right? I'm not really familiar with how this usually works 😅

I'm merging your PR now. Typically, there is nothing more to do for the contributor once a PR is approved and no additional feedback is requested.

PR #1183, opened 2025-05-05
Add ConcurrentHashMap.KeySetView serializer

bergander:
This PR adds support for serializing sets created by `ConcurrentHashMap.newKeySet()`.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
+++ b/src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
@@ -855,18 +855,18 @@ public class DefaultSerializers {
 
 	/** Serializer for {@link ConcurrentHashMap.KeySetView}.
 	 * @author Andreas Bergander */
-	public static class KeySetViewSerializer extends CollectionSerializer<ConcurrentHashMap.KeySetView> {
-		protected void writeHeader (Kryo kryo, Output output, ConcurrentHashMap.KeySetView set) {
+	public static class KeySetViewSerializer extends Serializer<ConcurrentHashMap.KeySetView> {
+		public void write (Kryo kryo, Output output, ConcurrentHashMap.KeySetView set) {
 			kryo.writeClassAndObject(output, set.getMap());
 			kryo.writeClassAndObject(output, set.getMappedValue());
 		}
 
-		protected ConcurrentHashMap.KeySetView create (Kryo kryo, Input input, Class<? extends ConcurrentHashMap.KeySetView> type, int size) {
+		public ConcurrentHashMap.KeySetView read (Kryo kryo, Input input, Class<? extends ConcurrentHashMap.KeySetView> type) {
 			return createKeySetView((ConcurrentHashMap)kryo.readClassAndObject(input), kryo.readClassAndObject(input));
 		}
 
-		protected ConcurrentHashMap.KeySetView createCopy (Kryo kryo, ConcurrentHashMap.KeySetView original) {
-			return createKeySetView(original.getMap(), original.getMappedValue());
+		public ConcurrentHashMap.KeySetView copy (Kryo kryo, ConcurrentHashMap.KeySetView original) {
+			return createKeySetView(kryo.copy(original.getMap()), kryo.copy(original.getMappedValue()));
 		}
 
 		private ConcurrentHashMap.KeySetView createKeySetView (ConcurrentHashMap map, Object mappedValue) {
--- a/src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
+++ b/src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
@@ -62,6 +62,7 @@ import java.util.TimeZone;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


src/com/esotericsoftware/kryo/Kryo.java
```diff
@@ -225,6 +227,7 @@ public Kryo (ClassResolver classResolver, ReferenceResolver referenceResolver) {
 		addDefaultSerializer(Arrays.asList().getClass(), ArraysAsListSerializer.class);
 		addDefaultSerializer(void.class, new VoidSerializer());
 		addDefaultSerializer(PriorityQueue.class, new PriorityQueueSerializer());
+		addDefaultSerializer(ConcurrentHashMap.KeySetView.class, new KeySetViewSerializer());
```
theigl, line 230: Please remove the serializer registration. It breaks backwards compatibility. If we add the serializer to Kryo 5, users will have to register it manually. It can only be registered as a default serializer in the next major version.

test/com/esotericsoftware/kryo/serializers/CollectionSerializerTest.java
```diff
@@ -112,6 +114,17 @@ void testCopy () {
 		assertNotSame(objects1.get(0), objects2.get(0));
 	}
 
+	@Test
+	void testConcurrentHashMapKeySetView() {
+		ConcurrentHashMap.KeySetView set = ConcurrentHashMap.newKeySet();
```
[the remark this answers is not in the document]
bergander: Yes, good point! I've updated the serializer to include the `mappedValue` and added test cases for it.

src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
```diff
@@ -852,6 +853,27 @@ private PriorityQueue createPriorityQueue (Class<? extends Collection> type, int
 		}
 	}
 
+	/** Serializer for {@link ConcurrentHashMap.KeySetView}.
+	 * @author Andreas Bergander */
+	public static class KeySetViewSerializer extends CollectionSerializer<ConcurrentHashMap.KeySetView> {
+		protected void writeHeader (Kryo kryo, Output output, ConcurrentHashMap.KeySetView set) {
+			kryo.writeClassAndObject(output, set.getMappedValue());
```
theigl, line 860: Is there a reason for writing the `mappedValue` first? If not, then please write the map first and change the order in `createKeySetView` as well.

Apart from that LGTM.

bergander: I've swapped the order of those fields now.

src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
```diff
@@ -852,6 +853,27 @@ private PriorityQueue createPriorityQueue (Class<? extends Collection> type, int
 		}
 	}
 
+	/** Serializer for {@link ConcurrentHashMap.KeySetView}.
+	 * @author Andreas Bergander */
+	public static class KeySetViewSerializer extends CollectionSerializer<ConcurrentHashMap.KeySetView> {
+		protected void writeHeader (Kryo kryo, Output output, ConcurrentHashMap.KeySetView set) {
```
theigl, line 859: I just realized that this is not ideal. We are writing the keys twice. Once here as part of the map in the header and then again in the main `write` method of the `CollectionSerializer`. It would be enough to write the map because the keyset is only a view. Can you try to implement this without extending `CollectionSerializer` and implement `Serializer` directly? This should make the serialized payload significantly smaller.

src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
```diff
@@ -852,6 +853,27 @@ private PriorityQueue createPriorityQueue (Class<? extends Collection> type, int
...
+			kryo.writeClassAndObject(output, set.getMap());
+			kryo.writeClassAndObject(output, set.getMappedValue());
+		}
+
+		protected ConcurrentHashMap.KeySetView create (Kryo kryo, Input input, Class<? extends ConcurrentHashMap.KeySetView> type, int size) {
+			return createKeySetView((ConcurrentHashMap)kryo.readClassAndObject(input), kryo.readClassAndObject(input));
+		}
+
+		protected ConcurrentHashMap.KeySetView createCopy (Kryo kryo, ConcurrentHashMap.KeySetView original) {
+			return createKeySetView(original.getMap(), original.getMappedValue());
```
theigl, line 869: I think we should deep copy here. Can you try something like this?

```java
      ConcurrentHashMap newMap = kryo.copy(value.getMap());
      return newMap.keySet(kryo.copy(value.getMappedValue()));
```

src/com/esotericsoftware/kryo/serializers/DefaultSerializers.java
theigl: `writeHeader` would simply become `write` and `create` would become `read`.

bergander: Thank you for pointing me in the right direction! PR is updated according to your comments.

theigl (approved):
That looks great now, thanks!

mrniko:
@theigl 

Is there expected release date for this?

theigl:
@mrniko I plan to do a release in the coming weeks.

PR #1220, opened 2025-10-03
Fix nondeterministic list comparison in doAssertEquals for PriorityQueue tests

yonghanlin:
A failure occurred in ```testPriorityQueueSubclass()``` and ```testPriorityQueue()``` in both the main and main-versioned modules when using NonDex to run these commands:
```
mvn -pl main-versioned edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=com.esotericsoftware.kryo.serializers.DefaultSerializersTest#testPriorityQueueSubclass
```
```
mvn -pl main-versioned edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=com.esotericsoftware.kryo.serializers.DefaultSerializersTest#testPriorityQueue
```
```
mvn -pl main edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=com.esotericsoftware.kryo.serializers.DefaultSerializersTest#testPriorityQueueSubclass
```
```
mvn -pl main edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=com.esotericsoftware.kryo.serializers.DefaultSerializersTest#testPriorityQueue
```

**Reasons for the failure:**
A PriorityQueue in Java is backed by a binary heap, which guarantees the smallest element can be accessed quickly but does not guarantee iteration order. When converted to a list, two queues with the same elements can produce different sequences depending on their internal heap structure. The original doAssertEquals used assertEquals on the lists, which compared both order and contents, so the tests failed even though the queues contained the same elements.

**The Fix:**
Updated doAssertEquals to handle list comparison by first checking list sizes to ensure element counts match. Then verifying that each list contains all elements of the other, ignoring order. For non-list values, the original assertEquals is preserved.

### DIFF
--- a/test/com/esotericsoftware/kryo/KryoTestCase.java
+++ b/test/com/esotericsoftware/kryo/KryoTestCase.java
@@ -282,7 +282,18 @@ public abstract class KryoTestCase {
 	}
 
 	protected void doAssertEquals (Object object1, Object object2) {
-		assertEquals(arrayToList(object1), arrayToList(object2));
+		Object val1 = arrayToList(object1);
+		Object val2 = arrayToList(object2);
+
+		if (val1 instanceof List && val2 instanceof List) {
+			List<?> list1 = (List<?>) val1;
+			List<?> list2 = (List<?>) val2;
+			assertEquals(list1.size(), list2.size());
+			assertTrue(list1.containsAll(list2));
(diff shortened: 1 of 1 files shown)


theigl:
Thanks!

PR #1229, opened 2025-11-02
Refactor: replace generic RuntimeException with custom Serialization Exception

shrujal24:
Bug :
The AbstractConcurrencyState.serialize() method was throwing a generic RuntimeException when an IOException occurred during serialization. This made it difficult to identify the root cause and violated clean code and SonarCloud recommendations for using specific or custom exceptions.

To Reproduce
Run any Kryo serialization benchmark (e.g., ConcurrencyBenchmark).
Trigger an I/O failure during the serialize() process (e.g., invalid output stream).
Observe that the code throws a generic RuntimeException without meaningful context.

Expected behavior:
A meaningful, domain-specific exception should be thrown to clearly indicate that the failure occurred during serialization.


theigl:
The benchmark code is basically test code. It is not part of the Kryo API and the type of exception thrown does not matter in this case.

PR #1283, opened 2026-06-17
Fix generics stack leak that poisoned reused Kryo instances

### LINKED ISSUE #1281: Exception during deserialization leaks generics stack; reset() doesn't clear it, poisoning a reused Kryo instance
## Summary

When an exception is thrown **while reading a field that carries generic type information**, the generic type that was pushed onto `DefaultGenerics` is never popped (the `pushGenericType` / `popGenericType` pair is not exception-safe). `Kryo.reset()` does **not** clear the generics stack, so even with `autoReset = true` the leaked entries survive. If the same `Kryo` instance is reused afterwards (the recommended thread-local / `Pool` usage), a later, completely valid and unrelated deserialization fails with a misleading `ArrayIndexOutOfBoundsException`. The instance is effectively poisoned for its remaining lifetime.

This is especially nasty because:

- the second failure happens on **good** data, far away from the original cause, so it looks like data corruption rather than a stale-state bug;
- the error type drifts (the first failure is e.g. "Unable to find class", the recurring failure is `ArrayIndexOutOfBoundsException`);
- it survives `reset()` / `Pool.free()`, so neither auto-reset nor pooling recovers the instance.

## Version

- Kryo **5.6.2** (latest release at time of writing)
- `references = false`, `registrationRequired = false`, default `autoReset = true`

## Minimal reproduction

```java
import com.esotericsoftware.kryo.Kryo;
import com.esotericsoftware.kryo.Serializer;
import com.esotericsoftware.kryo.io.Input;
import com.esotericsoftware.kryo.io.Output;
import java.util.*;

public class GenericsLeakRepro {

    public static class Holder { public List<Item> items; }
    public static class Item   { public Object payload; }
    public static class Payload {}

    // Stands in for ANY read-time failure (missing class, IOException, malformed data, ...).
    static class ThrowingSerializer extends Serializer<Payload> {
        public void write(Kryo k, Output o, Payload p) {}
        public Payload read(Kryo k, Input i, Class<? extends Payload> t) {
            throw new RuntimeException("simulated read-time failure");
        }
    }

    static Kryo newKryo() {
        Kryo k = new Kryo();
        k.setRegistrationRequired(false);
        k.setReferences(false);
        return k;
    }

    public static void main(String[] args) {
        Kryo writer = newKryo();

        Holder holder = new Holder();
        holder.items = new ArrayList<>();
        Item item = new Item();
        item.payload = new Payload();
        holder.items.add(item);
        Output o1 = new Output(4096);
        writer.writeClassAndObject(o1, holder);
        byte[] failingBytes = o1.toBytes();

        writer.reset();

        // A plain HashMap with two non-empty nested maps.
        Map<String, Object> map = new HashMap<>();
        Map<String, String> n1 = new HashMap<>(); n1.put("a", "1");
        Map<String, String> n2 = new HashMap<>(); n2.put("b", "2");
        map.put("k1", n1);
        map.put("k2", n2);
        Output o2 = new Output(4096);
        writer.writeClassAndObject(o2, map);
        byte[] plainMapBytes = o2.toBytes();

        // One reusable reader (e.g. a thread-local or pooled Kryo).
        Kryo reader = newKryo();
        reader.register(Payload.class, new ThrowingSerializer());

        // 1) A deserialization that fails inside a generic field.
        try {
            reader.readClassAndObject(new Input(failingBytes));
        } catch (Exception expected) {
            System.out.println("[1] first read failed as expected: " + expected.getClass().getSimpleName());
        }
        System.out.println("[1] generics stack size after reset = "
                + reader.getGenerics().getGenericTypesSize() + " (expected 0)");

        // 2) A totally unrelated, well-formed HashMap now fails on the SAME instance.
        try {
            Object result = reader.readClassAndObject(new Input(plainMapBytes));
            System.out.println("[2] second read OK: " + result);
        } catch (Throwable t) {
            System.out.println("[2] second read FAILED on a valid payload: " + t);
        }
    }
}
```

### Actual output

```
[1] first read failed as expected: KryoException
[1] generics stack size after reset = 2 (expected 0)
[2] second read FAILED on a valid payload: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
```

### Expected

- After the first (failed) `readClassAndObject` returns, `getGenerics().getGenericTypesSize()` should be `0` — the generics stack should be empty once the object graph is torn down, exactly as the README states ("object graph state is typically reset automatically after each serialization").
- The second `readClassAndObject` on a valid payload should succeed and return the `HashMap`.

## Root cause (pointers)

1. `ReflectField.read(...)` calls `kryo.getGenerics().pushGenericType(...)` before reading the field value and `popGenericType()` after — but the `pop` is **not** in a `finally`, so any exception from the nested read skips it.
2. `Kryo.reset()` resets references / graph context / class resolver but **does not touch `generics`**, so the leaked entries persist across `readClassAndObject` calls (and across `Pool.free()` / `obtain()`).
3. On the next read, `nextGenericTypes()` matches a stale entry by depth only, hands a `List`-shaped (length-1) generic array to a `Map` read path, which indexes `[1]` → `ArrayIndexOutOfBoundsException`.

The same exception-safety gap exists in other push/pop sites (e.g. `MapSerializer` / `CollectionSerializer`).

## Suggested fix

Either (or both):

- Have `Kryo.reset()` also clear the generics stack (`generics.reset()` / pop-all), so auto-reset and pooling fully recover the instance; and/or
- Make the `pushGenericType` / `popGenericType` pairs exception-safe by moving the `pop` into a `finally` block at each call site.

Happy to send a PR if a preferred direction is indicated.

Icesource:
Fixes #1281.

## Problem

When a serializer throws while (de)serializing a field, entries are left on Kryo's generics stacks:
- `ReflectField` skips its `popGenericType` (it was after the nested read/write, not in a `finally`).
- `FieldSerializer` skips `popTypeVariables` (it is after the field loop, outside the `try`/`catch`).

`Kryo.reset()` did not clear these stacks, so the leaked state survives `autoReset` and poisons a reused (thread-local or pooled) `Kryo` instance — a later, unrelated (de)serialization then mistakes the stale generic for its own and fails with a misleading `ArrayIndexOutOfBoundsException`. The instance stays broken for the rest of its lifetime.

## Fix

Following the approach suggested in the issue:

1. **`Generics.reset()`**, called from `Kryo.reset()` — centralized so it always runs, even if a `finally` is forgotten somewhere. It has a fast path that returns immediately when the stacks are already empty (the normal case after a balanced serialization), so the per-serialization cost is negligible. The slow path nulls out the leaked entries and zeroes the sizes. Both the `genericTypes` stack and the type-variable (`arguments`) stack are cleared, since both leak on exception. `NoGenerics` gets a no-op.
2. **`popGenericType` moved to a `finally`** in `ReflectField.read`/`write` — the most likely source of serialization exceptions — so the common case keeps the stack balanced without relying on `reset()`. `popGenericType` is already safe to call when nothing was pushed.

## Tests

`GenericsResetTest` covers both directions end-to-end:
- **Deserialization**: a read that throws inside a `List<Item>` field, followed by an unrelated `HashMap` with two non-empty nested maps. Before the fix the generics stack is left at size 2 and the second read throws `ArrayIndexOutOfBoundsException`; after the fix the stack is empty and the second read succeeds.
- **Serialization**: a write that throws inside the same field, then asserting the stack is empty and a subsequent serialization/deserialization succeeds.

Both fail on `master` without the fix and pass with it.

Full suite: the only failure is `DefaultSerializersTest#testLocaleSerializer` (`zh_CN` vs `zh_CN_#Hans`), a pre-existing locale-data difference in my environment that also fails on unmodified `master`.

### DIFF
--- a/src/com/esotericsoftware/kryo/Kryo.java
+++ b/src/com/esotericsoftware/kryo/Kryo.java
@@ -953,6 +953,7 @@ public class Kryo {
 		depth = 0;
 		if (graphContext != null) graphContext.clear(2048);
 		classResolver.reset();
+		generics.reset();
 		if (references) {
 			referenceResolver.reset();
 			readObject = null;
--- a/src/com/esotericsoftware/kryo/serializers/ReflectField.java
+++ b/src/com/esotericsoftware/kryo/serializers/ReflectField.java
@@ -86,7 +86,6 @@ class ReflectField extends CachedField {
 					kryo.writeObject(output, value, serializer);
 				}
 			}
-			kryo.getGenerics().popGenericType();
 		} catch (IllegalAccessException ex) {
 			throw new KryoException("Error accessing field: " + name + " (" + object.getClass().getName() + ")", ex);
 		} catch (KryoException ex) {
@@ -102,6 +101,9 @@ class ReflectField extends CachedField {
 			KryoException ex = new KryoException(t);
 			ex.addTrace(name + " (" + object.getClass().getName() + ")");
 			throw ex;
+		} finally {
+			// Pop in a finally so an exception thrown by the nested write does not leave the generics stack unbalanced.
+			kryo.getGenerics().popGenericType();
 		}
 	}
 
@@ -134,8 +136,6 @@ class ReflectField extends CachedField {
 				else
 					value = kryo.readObject(input, concreteType, serializer);
 			}
-			kryo.getGenerics().popGenericType();
-
 			set(object, value);
 		} catch (IllegalAccessException ex) {
 			throw new KryoException("Error accessing field: " + name + " (" + fieldSerializer.type.getName() + ")", ex);
@@ -146,6 +146,9 @@ class ReflectField extends CachedField {
 			KryoException ex = new KryoException(t);
 			ex.addTrace(name + " (" + fieldSerializer.type.getName() + ")");
 			throw ex;
+		} finally {
+			// Pop in a finally so an exception thrown by the nested read does not leave the generics stack unbalanced.
+			kryo.getGenerics().popGenericType();
 		}
 	}
 
--- a/src/com/esotericsoftware/kryo/util/DefaultGenerics.java
+++ b/src/com/esotericsoftware/kryo/util/DefaultGenerics.java
@@ -156,6 +156,19 @@ public final class DefaultGenerics implements Generics {
 		return genericTypesSize;
 	}
 
+	@Override
+	public void reset () {
+		// Fast path: after a balanced (successful) serialization both stacks are already empty.
+		if (genericTypesSize == 0 && argumentsSize == 0) return;
+		// Slow path: a serializer threw before popping. Discard the leaked entries so the next serialization starts clean.
+		for (int i = 0; i < genericTypesSize; i++)
+			genericTypes[i] = null;
+		genericTypesSize = 0;
+		for (int i = 0; i < argumentsSize; i++)
+			arguments[i] = null;
+		argumentsSize = 0;
+	}
+
 	public String toString () {
 		StringBuilder buffer = new StringBuilder();
 		for (int i = 0; i < argumentsSize; i += 2) {
--- a/src/com/esotericsoftware/kryo/util/Generics.java
+++ b/src/com/esotericsoftware/kryo/util/Generics.java
@@ -21,6 +21,8 @@ package com.esotericsoftware.kryo.util;
 
 import static com.esotericsoftware.kryo.util.Util.*;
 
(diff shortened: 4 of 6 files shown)


Icesource:
Thanks for the review. Addressed both points:

- Made `Generics.reset()` a `default` no-op method so it is not a breaking change for external `Generics` implementations; `DefaultGenerics` still overrides it with the real reset.
- Updated the Javadoc to note it runs after both serialization and deserialization (when auto-reset is enabled).

theigl:
That looks great! Thank you @Icesource.

Issue #1282, opened 2026-06-16
Unbounded new int[length] from declared array length in Input.readInts → OutOfMemoryError on malformed input


joszamama:
## Describe the bug
When reading a primitive array, Kryo allocates the array from the *declared* length before validating it against the bytes actually remaining in the input. A few-byte input that declares a huge length causes an immediate `OutOfMemoryError` (`new int[length]` with no bound). `Input.readInts` and its siblings (`readLongs`/`readFloats`/`readDoubles`/...) and the array/collection serializers share this pattern.

Reporting this as a robustness/hardening issue. I am aware Kryo is not intended for deserializing untrusted data; filing in case bounding the initial allocation is considered worthwhile.

## To Reproduce
Minimal runnable example (5-byte input):

  ```java
  import com.esotericsoftware.kryo.Kryo;
  import com.esotericsoftware.kryo.io.*;

  public class KryoPoc {
    public static void main(String[] a) {
      Kryo kryo = new Kryo();
      kryo.register(int[].class);          // typical application registration
      Output o = new Output(16);
      o.writeVarInt(2000000001, true);     // length+1 -> declared int[] length 2,000,000,000
      o.flush();
      byte[] evil = o.toBytes();           // 5 bytes; no element payload follows
      kryo.readObject(new Input(evil), int[].class);   // OutOfMemoryError here
    }
  }
  ```
  
  Result:
  ```
  Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
      at com.esotericsoftware.kryo.io.Input.readInts(Input.java:968)
  ```
The input declares a 2,000,000,000-element array but carries no element data; Kryo allocates `new int[2000000000]` before checking that the input could possibly contain that many elements.

## Environment:

   - OS: macOS 26.5.1 (25F80)
   - JDK Version: 21
   - Kryo Version: 5.6.2 (latest)

## Suggested fix 
Bound the initial allocation to the bytes remaining in the input (a primitive element is at least 1 byte, so the declared length cannot exceed the remaining byte count), or read and grow in chunks rather than pre-allocating the full declared length. Found via boundary-value testing as part of academic research.

theigl:
Thanks for the report. This is a known limitation of Kryo, which assumes that data comes from a trusted source. 

Implementing some kind of reasonable guardrails for deserialization would definitely make sense. Not only for arrays and collections, but also for Strings.

[...]

PR #1284, opened 2026-06-18
Fix for Issue #1282: Guard deserialization against unbounded allocation from declared sizes

joszamama:
## Summary

Several serializers read a declared length, count, or size and allocate from it before validating it against the available input. A small corrupt or malicious message that declares a multi-billion element array, string, collection, or map triggers a multi-gigabyte allocation and an `OutOfMemoryError`. This continues the discussion in #1282 (primitive arrays) and extends the same guard to the other read paths that share the pattern, plus an opt-in size limit for callers that decode untrusted input.

Kryo assumes a trusted source, so these are robustness fixes, not a change to that threat model. The default behavior is unchanged.

## The pattern

`new T[declaredLength]` (and `new char[charCount]`, `new ArrayList<>(declaredSize)`, `new HashMap(declaredSize)`, ...) runs before any element is read, so the declared size alone drives the allocation. A 6 to 10 byte message can request gigabytes.

## What this changes

When an `Input` is backed by a fixed buffer (no `InputStream`), the number of elements cannot exceed the number of bytes remaining, because every element occupies at least one byte. That is the bound used here.

- **Primitive arrays and `byte[]`** (`Input.readInts/readLongs/.../readBytes`, and the `ByteBufferInput`, `UnsafeInput`, `UnsafeByteBufferInput` variants): validate the declared length against the bytes remaining before allocating. This is the #1282 case.
- **Strings** (`readUtf8Chars` in `Input` and `ByteBufferInput`): the same guard on the UTF-8 char count.
- **`String[]`, closure captured args, and `CompatibleFieldSerializer` field counts**: the same guard. Each element is read as a string or a class-tagged object, so it is at least one byte.
- **Collections and Maps** (`CollectionSerializer`, `MapSerializer`, including the `Arrays.asList`, JDK immutable, and `PriorityQueue` create overrides): clamp the initial capacity to the bytes remaining instead of throwing. The collection or map still grows to the real element count, so valid input is never rejected, while a corrupt size no longer forces a large pre-allocation. The clamp is applied in `read()` before `create()` so the create overrides are covered too.

## Why throw in some places and clamp in others

Throwing is only safe where every element is guaranteed to consume at least one byte (primitives, bytes, chars, strings, class-tagged objects). There, a declared length greater than the bytes remaining is provably malformed, so no valid input is rejected.

Collections and maps can hold elements that serialize to zero bytes (for example a registered no-field element type written with a known serializer), so a valid collection can legitimately declare more elements than there are bytes remaining at that point. Throwing there would reject valid input. Clamping the capacity avoids that: it caps the pre-allocation but lets the structure grow, so the result is always correct.

## What is intentionally left unchanged

`ObjectArraySerializer` (generic `Object[]`) is not guarded by the bytes-remaining check. Its elements can serialize to zero bytes and the array is fixed size, so neither the throw nor the clamp can bound it without rejecting valid input. The `maxArraySize` limit below still applies to it.

## maxArraySize: the configurable limit (default off), and why

The bytes-remaining guard only applies to a fixed buffer. For an `InputStream`-backed `Input` the buffered window is not the total size, so a declared size cannot be checked against it without rejecting valid large input. To let callers bound streaming (and buffered) deserialization, this adds:

```java
input.setMaxArraySize(n); // default Integer.MAX_VALUE
```

`validateArrayLength` and the collection and map size reads reject any declared size above the limit, on both stream-backed and buffer-backed input.

The default is `Integer.MAX_VALUE`, that is, unlimited:

- **By default nothing changes.** A valid payload always has length <= bytes remaining, so the
 buffer guard never fires on valid input, and the unlimited cap never fires at all. Existing callers see today's behavior, and the in-memory case is still protected automatically.
- **Callers that decode untrusted data opt in** with a value suited to their application. This is the only mechanism that covers stream-backed input, which cannot be bounded automatically.

A finite default was considered but not chosen. A finite default would reject legitimately large payloads from existing users and would change behavior on upgrade. Keeping it unlimited preserves backward compatibility and matches Kryo's trusted-source model: a caller that exposes Kryo to untrusted input sets the bound it expects.

## Backwards compatibility

- Default behavior is unchanged. Large valid arrays, strings, collections, and maps still deserialize. The guard rejects only a declared size larger than the input can supply.
- The one observable change for malformed input is the exception type: `KryoBufferUnderflowException` or `KryoException` instead of `OutOfMemoryError` or `NegativeArraySizeException`. It is catchable and thrown before the allocation.
- Stream-backed input is unchanged unless `setMaxArraySize` is called.

## New tests

- `InputOutputTest.testArrayLengthValidation`, `testMaliciousArrayLength`: the primitive array and `byte[]` guards, including an end-to-end `int[]` decode.
- `InputOutputTest.testMaliciousStringLength`: the string char-count guard on `Input` and `ByteBufferInput`.
- `InputOutputTest.testMaxArraySize`: the limit on a stream-backed input for arrays.
- `UnsafeInputOutputTest.testArrayLengthValidation`: the unsafe array variants.
- `CollectionSerializerTest.testMaliciousCollectionSize`, `testMaxCollectionSize`: the collection clamp and the stream limit.
- `MapSerializerTest.testMaliciousMapSize`, `testMaxMapSize`: the map clamp and the stream limit.

## Verification

The full module test suite passes. It was also run on the project's JDK matrix (Temurin 8, 11, 17, 21, 25), building on 11 and testing on each, as in CI.

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
@@ -172,6 +172,10 @@ public class MapSerializer<T extends Map> extends Serializer<T> {
 	/** Used by {@link #read(Kryo, Input, Class)} to create the new object. This can be overridden to customize object creation, eg
 	 * to call a constructor with arguments. The default implementation uses {@link Kryo#newInstance(Class)} with a special case
 	 * for HashMap. */
+	private static int clampSize (Input input, int size) {
+		return input.getInputStream() == null ? Math.min(size, input.limit() - input.position()) : size;
+	}
+
 	protected T create (Kryo kryo, Input input, Class<? extends T> type, int size) {
 		if (type == HashMap.class) {
 			if (size < 3)
@@ -188,7 +192,7 @@ public class MapSerializer<T extends Map> extends Serializer<T> {
 		if (length == 0) return null;
 		length--;
 
-		T map = create(kryo, input, type, length);
+		T map = create(kryo, input, type, clampSize(input, length));
 		kryo.reference(map);
 		if (length == 0) return map;
 
--- a/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
@@ -175,7 +175,7 @@ public class MapSerializer<T extends Map> extends Serializer<T> {
 	 * for HashMap. */
 	private static int clampSize (Input input, int size) {
 		if (size > input.getMaxArraySize())
-			throw new KryoException("Array size is larger than maxArraySize: " + size + " > " + input.getMaxArraySize());
+			throw new KryoException("Declared size larger than maxArraySize: " + size + " > " + input.getMaxArraySize());
 		return input.getInputStream() == null ? Math.min(size, input.limit() - input.position()) : size;
 	}
 
--- a/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
@@ -20,7 +20,6 @@
 package com.esotericsoftware.kryo.serializers;
 
 import com.esotericsoftware.kryo.Kryo;
-import com.esotericsoftware.kryo.KryoException;
 import com.esotericsoftware.kryo.Serializer;
 import com.esotericsoftware.kryo.SerializerFactory;
 import com.esotericsoftware.kryo.io.Input;
@@ -173,12 +172,6 @@ public class MapSerializer<T extends Map> extends Serializer<T> {
 	/** Used by {@link #read(Kryo, Input, Class)} to create the new object. This can be overridden to customize object creation, eg
 	 * to call a constructor with arguments. The default implementation uses {@link Kryo#newInstance(Class)} with a special case
 	 * for HashMap. */
-	private static int clampSize (Input input, int size) {
-		if (size > input.getMaxArraySize())
-			throw new KryoException("Declared size larger than maxArraySize: " + size + " > " + input.getMaxArraySize());
-		return input.getInputStream() == null ? Math.min(size, input.limit() - input.position()) : size;
-	}
-
 	protected T create (Kryo kryo, Input input, Class<? extends T> type, int size) {
 		if (type == HashMap.class) {
 			if (size < 3)
@@ -195,7 +188,7 @@ public class MapSerializer<T extends Map> extends Serializer<T> {
 		if (length == 0) return null;
 		length--;
 
-		T map = create(kryo, input, type, clampSize(input, length));
+		T map = create(kryo, input, type, input.clampSize(length));
 		kryo.reference(map);
 		if (length == 0) return map;
 
--- a/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/MapSerializer.java
@@ -20,6 +20,7 @@
 package com.esotericsoftware.kryo.serializers;
 
 import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.KryoException;
 import com.esotericsoftware.kryo.Serializer;
 import com.esotericsoftware.kryo.SerializerFactory;
 import com.esotericsoftware.kryo.io.Input;
@@ -173,6 +174,8 @@ public class MapSerializer<T extends Map> extends Serializer<T> {
 	 * to call a constructor with arguments. The default implementation uses {@link Kryo#newInstance(Class)} with a special case
 	 * for HashMap. */
 	private static int clampSize (Input input, int size) {
+		if (size > input.getMaxArraySize())
+			throw new KryoException("Array size is larger than maxArraySize: " + size + " > " + input.getMaxArraySize());
 		return input.getInputStream() == null ? Math.min(size, input.limit() - input.position()) : size;
 	}
 
--- a/src/com/esotericsoftware/kryo/unsafe/UnsafeByteBufferInput.java
+++ b/src/com/esotericsoftware/kryo/unsafe/UnsafeByteBufferInput.java
@@ -185,37 +185,37 @@ public class UnsafeByteBufferInput extends ByteBufferInput {
 	}
 
 	public int[] readInts (int length) throws KryoException {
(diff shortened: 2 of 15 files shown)


theigl:
@joszamama: This looks great already.

Except for the issues raised by Copilot (from a quick glance, most of them seem valid), I would ask you for two more changes:

1. Document the new configuration option for `Input` in the readme and add javadoc for the option.
2. Add a benchmark for array deserialization to `kryo-benchmarks` and check if there are any performance regressions. My guess would be that the overhead is negligible, but it would be great to know for sure.

It might make sense to separate array and collection `maxArraySize` from the max size for strings. We could configure reasonable defaults for the next major release and the value for `maxArraySize` could probably be much lower than the one for strings. What do you think?

theigl:
> It might make sense to separate array and collection maxArraySize from the max size for strings. We could configure reasonable defaults for the next major release and the value for maxArraySize could probably be much lower than the one for strings. What do you think?

Alternatively, we can start out with the single configuration option, keep it as simple as possible and split it later as needed.

joszamama:
Thanks @theigl. Pushed a follow-up addressing the review.

**Copilot points**

- The fixed-width array readers (`readInts`, `readLongs`, `readFloats`, `readDoubles`, `readShorts`, `readChars`) now validate by byte count instead of element count. I added a `validateArrayLength(int length, int bytesPerElement)` overload and use it in `Input`, `ByteBufferInput`, `UnsafeInput` and `UnsafeByteBufferInput`. It rejects a declared size whose total byte size is larger than the bytes remaining (buffer-backed) or overflows an int (stream-backed), so the `length << n` shift can no longer overflow. The one-byte paths (byte arrays, UTF-8, var-length, objects) keep the existing check.
- `setMaxArraySize` now rejects negative values with an `IllegalArgumentException`.
- Renamed the message from "Array size" to "Declared size" in `Input`, `CollectionSerializer` and `MapSerializer`, since it also fires for strings, collections and maps.

**Docs**

Added a "Limiting deserialized size" section to the README and javadoc on `getMaxArraySize`, `setMaxArraySize` and both `validateArrayLength` overloads.

**Benchmark**

Added `readDoubles`/`writeDoubles` to `ArrayBenchmark` (it already covered ints and longs) and ran the array read benchmarks with and without the guard to isolate its cost. The overhead is within noise:

| array read | unguarded | guarded |
| --- | --- | --- |
| readInts | 0.230 ± 0.074 | 0.228 ± 0.065 |
| readLongs | 0.514 ± 0.082 | 0.513 ± 0.086 |
| readDoubles | 0.511 ± 0.086 | 0.505 ± 0.098 |

(s/op, SingleShotTime, batchSize 12M, single fork, local JDK 24 on macOS arm64.) The check runs once per array read.

joszamama:
> > It might make sense to separate array and collection maxArraySize from the max size for strings. We could configure reasonable defaults for the next major release and the value for maxArraySize could probably be much lower than the one for strings. What do you think?
> 
> Alternatively, we can start out with the single configuration option, keep it as simple as possible and split it later as needed.

I'd keep the single option for now. Splitting strings from arrays and collections fits better together with choosing finite defaults, which is a behavior change (hence, better scheduled for the next major release?) I am more in the security side than on the maintenance side, so I am sure your opinion would be more valuable.

src/com/esotericsoftware/kryo/serializers/MapSerializer.java
```diff
@@ -172,6 +173,12 @@ protected void writeHeader (Kryo kryo, Output output, T map) {
 	/** Used by {@link #read(Kryo, Input, Class)} to create the new object. This can be overridden to customize object creation, eg
 	 * to call a constructor with arguments. The default implementation uses {@link Kryo#newInstance(Class)} with a special case
 	 * for HashMap. */
+	private static int clampSize (Input input, int size) {
```
theigl: The JavaDoc belongs to the create method. Please move it there.

Also move `clampSize` to Input (instance method) and re-use it between Map and Collection serializers.

src/com/esotericsoftware/kryo/unsafe/UnsafeByteBufferInput.java
```diff
@@ -185,43 +185,43 @@ public boolean readBoolean () throws KryoException {
 	}
 
 	public int[] readInts (int length) throws KryoException {
-		int[] array = new int[length];
+		int[] array = new int[validateArrayLength(length, 4)];
```
theigl: Please use constants `Integer.BYTES`, `Float.BYTES`, `Long.BYTES` etc.

theigl (commented):
Looks great! Only two minor comments.

I'll let Copilot review once more.

joszamama:
Thanks for the review! Addressed all three:
  
  - Moved `clampSize` into `Input` as an instance method and reused it from both the Map and Collection serializers (dropped the duplicated copies).
  - Switched to `Integer.BYTES` / `Long.BYTES` / `Float.BYTES` / `Double.BYTES` / `Short.BYTES` / `Character.BYTES` instead of numeric literals, both for the `validateArrayLength` element widths and for the unsafe bulk-read byte counts (the old `length << n`). That also covers the earlier Copilot note about fixed-width reads.
  - Moved the `create()` JavaDoc onto `create` in `MapSerializer`, and removed the now-unused `KryoException` imports.

All 325 core-module tests still pass.

theigl:
Thank you very much for the contribution @joszamama! 

I had guardrails on the back of my mind for quite a while, but never found the time to work on them, so this is highly appreciated.

xingchen20lj:
Thanks for merging #1284 — the unbounded-allocation guard looks like the
right fix. Two quick questions since the latest release (5.6.2, Oct 2024)
still contains the vulnerable pattern:

1. Does the Input.java change in this PR cover the readString ->
   readUtf8Chars path (`new char[charCount]` with no upper bound)? From the
   diff it looks like it does, but I'd like to confirm.
2. Is there a rough timeline for the next release (5.6.3+) so this fix
   actually ships?

Context: with default configuration, `readClassAndObject` reads the class
name via `readString` before the registrationRequired guard, so a crafted
stream declaring charCount=0x7FFFFFFF triggers OutOfMemoryError on the
current release. Happy to share a minimal repro / verification matrix if
useful. Thanks for maintaining Kryo!

theigl:
@xingchen20lj:

1. Yes, String reading is covered.
2. I'm planning to do a 5.7 release in September.

PR #1295, opened 2026-08-03
Add an optional class filter to JavaSerializer deserialization

Nexory:
## What / Why

`JavaSerializer.read(...)` deserializes with `ObjectInputStream.readObject()`, and its `ObjectInputStreamWithKryoClassLoader` overrides only `resolveClass(...)` for classloader resolution. It applies no restriction on which classes may be deserialized. Kryo's own `setRegistrationRequired(true)` does not help on this path, because the bytes are handed to the JDK serialization mechanism rather than resolved through Kryo's class registration, so a Kryo stream that uses `JavaSerializer` for any field type deserializes arbitrary classes from the input.

This adds an opt-in class filter to `JavaSerializer` as defense-in-depth when reading serialized data from an untrusted source:

- `setClassFilter(Predicate<Class>)` / `getClassFilter()`.
- The filter is checked in `resolveClass(...)`: a class for which the predicate returns `false` is rejected with a `KryoException` before it is used.

## Non-breaking

The default is unchanged. With no filter set (`null`, the default) behavior is exactly as before, and the existing classloader-resolution logic in `resolveClass(...)` is preserved.

## Java 8 note

Kryo targets Java 8, so this uses a `Predicate<Class>` checked in the existing `resolveClass(...)` hook rather than the Java 9+ `java.io.ObjectInputFilter` (JEP-290). If you prefer, I am happy to additionally wire a real `ObjectInputFilter` on Java 9+ (resolved reflectively so the Java 8 build is unaffected), or to expose a `Set<String>` allowlist instead of a predicate. Whichever shape you prefer, the check stays in one place.

## Tests

`JavaSerializerTest` adds two cases: a filter that disallows the class rejects deserialization (as `KryoException`), and a filter that allows it still round-trips. The existing tests are unchanged. Verified with `mvn -Dtest=JavaSerializerTest test` on JDK 11 (source/target 8).

### DIFF
--- a/src/com/esotericsoftware/kryo/serializers/JavaSerializer.java
+++ b/src/com/esotericsoftware/kryo/serializers/JavaSerializer.java
@@ -32,6 +32,7 @@ import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.ObjectStreamClass;
+import java.util.function.Predicate;
 
 /** Serializes objects using Java's built in serialization mechanism. Note that this is very inefficient and should be avoided if
  * possible.
@@ -40,6 +41,21 @@ import java.io.ObjectStreamClass;
  * @see KryoSerializable
  * @author Nathan Sweet */
 public class JavaSerializer extends Serializer {
+	private Predicate<Class> classFilter;
+
+	/** Sets an optional filter applied to each class encountered while deserializing. When set, a class for which the predicate
+	 * returns false is rejected with a {@link KryoException} before it is used, providing opt-in, defense-in-depth protection when
+	 * reading serialized data from an untrusted source. When null (the default) no filtering is applied and behavior is unchanged.
+	 * @param classFilter May be null. */
+	public void setClassFilter (Predicate<Class> classFilter) {
+		this.classFilter = classFilter;
+	}
+
+	/** @return May be null. */
+	public Predicate<Class> getClassFilter () {
+		return classFilter;
+	}
+
 	public void write (Kryo kryo, Output output, Object object) {
 		try {
 			ObjectMap graphContext = kryo.getGraphContext();
@@ -60,7 +76,7 @@ public class JavaSerializer extends Serializer {
 			ObjectMap graphContext = kryo.getGraphContext();
 			ObjectInputStream objectStream = (ObjectInputStream)graphContext.get(this);
 			if (objectStream == null) {
-				objectStream = new ObjectInputStreamWithKryoClassLoader(input, kryo);
+				objectStream = new ObjectInputStreamWithKryoClassLoader(input, kryo, classFilter);
 				graphContext.put(this, objectStream);
 			}
 			return objectStream.readObject();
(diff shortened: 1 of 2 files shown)


theigl:
Hi @Nexory. Thanks for the PR. The issue I see is that your filter is applied to late. It does `Class.forName` and **then*** checks if the class is allowed. The potentially malicious class is already loaded at this point. The check would need to be done before class loading and match on the class name.

 Kryo 6 will move to Java 17 or 21 and we could replace this with an `ObjectInputFilter` then.

Nexory:
You are right that the check belongs before the class is resolved, and it now is: the filter
runs on `type.getName()` as the first thing `resolveClass` does, so a refused class is never
loaded and a name that does not resolve at all is refused rather than reported as missing. The
type changes from `Predicate<Class>` to `Predicate<String>` accordingly.

One correction to the reasoning, because it affects how urgent this was. The old code called
`Class.forName(name, false, loader)`, and that `false` means do not initialize. Measured with a
class whose static initializer prints: resolving it that way does not run the initializer,
while `Class.forName(name)` does. So the class was being loaded but no attacker code ran at
that point. The change is still the right one, for the reasons above.

Measured with `-verbose:class`, on a gadget nested inside an allowed outer type, since that is
where the filter applies: with the filter refusing, the class does not appear in the load list
and its initializer does not run; with the filter allowing, both happen. The reader never names
the class and the bytes come from a separate JVM.

Worth noting for the Kryo 6 plan: this is the same shape `ObjectInputFilter` uses, a decision
on the name before resolution, so swapping the implementation later should be small.

I followed your PR workflow: build on JDK 11, then `mvn -B test` on 8 and on 17. 310 and 330
tests, no failures. I did not run the 11, 21, 25 and 26 legs.

theigl:
> One correction to the reasoning, because it affects how urgent this was. The old code called
Class.forName(name, false, loader), and that false means do not initialize. Measured with a
class whose static initializer prints: resolving it that way does not run the initializer,
while Class.forName(name) does. So the class was being loaded but no attacker code ran at
that point. The change is still the right one, for the reasons above.

You are right! I overlooked the `false` parameter. Now I'm not so sure anymore what the ideal shape of the filter should be. `ObjectInputFilter` can be constructed from a String pattern as well as a `Predicate<Class>`, but the final check is always against a `FilterInfo` which exposes a `Class`. So arguably, a `Predicate<Class>` is closer to `ObjectInputFilter`.

It's unfortunate that we are still on JDK8 and can't go for `ObjectInputFilter` directly.

What's your opinion? Should we revert back to your original version, or should we go with `String`?

Nexory:
Good question, and it turned out to be measurable. I ran the JDK's own filter under
`-verbose:class`, same setup as before, a gadget nested inside an allowed outer type: with
`ObjectInputFilter` refusing it, the class is loaded anyway. The filter is handed a resolved
`Class`, so the load has already happened when it decides. It does not initialize, same as
before, but it is in the load list.

So you are right that `Predicate<Class>` mirrors `ObjectInputFilter` exactly. What follows from
the same measurement is that the `String` version is not merely a different shape, it is
stricter than what `ObjectInputFilter` gives you: with it the class never enters the load list
at all.

My opinion, for what it is worth: keep `String`. Deciding before the load is the point of a
deserialization filter, and it also lets a caller refuse a name that does not resolve. The
migration stays one line:
`info -> info.serialClass() == null || predicate.test(info.serialClass().getName())`. The null
guard is not decoration: the API returns null for the checks where no new object is created,
the ones that let a filter look at depth, references and stream size.

The honest cost is expressiveness. A name predicate cannot say "allow anything implementing X"
without loading, and `Class` can. If you expect users to want that, `Class` is the better API
and the weaker guarantee is a deliberate trade. Either way I am happy to switch it back.

Issue #1299, opened 2026-08-20
ExternalizableSerializer probes writeReplace()/readResolve() once per object instead of once per class


vpaturet:
**Describe the bug**
This is a performance issue (no functional impact) that affects the speed of deserializing an object graph containing `Externalizable` objects.

`ExternalizableSerializer` needs to know whether a type uses the `writeReplace()`/`readResolve() `hooks, since those types require the JavaSerializer fallback. The class has a cache field for that decision, but nothing ever writes to it, making the cache inert:

``` java
private ObjectMap<Class, JavaSerializer> javaSerializerByType;

private JavaSerializer getJavaSerializerIfRequired (Class type) {
    JavaSerializer javaSerializer = getCachedSerializer(type);
    if (javaSerializer == null && isJavaSerializerRequired(type)) javaSerializer = new JavaSerializer();
    return javaSerializer;                        // result is never stored
}

private JavaSerializer getCachedSerializer (Class type) {
    if (javaSerializerByType == null) {
        javaSerializerByType = new ObjectMap();
        return null;
    }
    return javaSerializerByType.get(type);        // map is always empty -> always a miss
}
```

Proposed fix: add the missing logic to populate the cache.

```java
	private JavaSerializer getJavaSerializerIfRequired (Class type) {
		if (javaSerializerByType == null) {
			javaSerializerByType = new ObjectMap();
		} else if (javaSerializerByType.containsKey(type)) {
			return javaSerializerByType.get(type);
		}
		JavaSerializer javaSerializer = isJavaSerializerRequired(type) ? new JavaSerializer() : null;
		javaSerializerByType.put(type, javaSerializer);
		return javaSerializer;
	}
```

**To Reproduce**
I can provide a reproducer and a PR.

**Environment:**
 - OS: Ubuntu
 - JDK Version: 25
 - Kryo Version: 5.5.0

**Additional context**
Found in [OpenTripPlanner](https://github.com/opentripplanner/OpenTripPlanner), which uses Kryo to
serialize its routing graph and registers `ExternalizableSerializer` for Trove's `TPrimitiveHash`
subtree.
The observed performance penalty is ~20% more time to deserialize the object graph.

theigl:
Good catch. Do you want to create a PR for this?

vpaturet:
Sure, here is the PR: #1300

PR #1300, opened 2026-08-20
Cache the Java serializer lookup per type

vpaturet:
This fixes issue #1299 by implementing the missing logic in the ExternalizableSerializer cache.


theigl:
I'm wondering if we should replace this whole lookup with a `ClassValue`.

```java
	private final ClassValue<JavaSerializer> javaSerializerByType = new ClassValue<>() {
		protected JavaSerializer computeValue(Class<?> type) {
			return isJavaSerializerRequired(type) ? new JavaSerializer() : null;
		}
	};
```

Then the whole method just becomes:

```java
	private JavaSerializer getJavaSerializerIfRequired (Class type) {
		return javaSerializerByType.get(type);
	}
```

This is faster than the map lookup and much cleaner.

vpaturet:
This looks much better! I added one more commit with your suggestion.

theigl:
Thank you @vpaturet!

Issue #1239, opened 2025-12-05
java.util.ImmutableCollections$ListN can contain null values


mjrosenb:
**Describe the bug**
In older java versions, `List::of` was the only way to create a java.util.ImmutableCollections$ListN object, but now `Stream::toList` is also capable of creating them, and it does not have the limitation of not containing nulls

**To Reproduce**
There are two slightly different issues. It should be possible to fix both at the same time.
  In this case,  immutableWithNulls only has null values in it, so kryo successfully serializes it, but then can not deserialize it because it ends up calling `List.of(null, null)` 
```  @Test
  public void testImmutableDeserialization() throws Exception {
    final Kryo kryo = new Kryo();
    kryo.setRegistrationRequired(false);
    final Output output = new Output(64);
    final List<Integer> immutableWithNulls = Stream.of(null,1, null).filter(x -> x == null || x > 10).toList();

    System.out.println(immutableWithNulls.getClass());
    kryo.writeClassAndObject(output, immutableWithNulls);
    try(Input input = new Input(output.getBuffer())) {
      final Object copy = kryo.readClassAndObject(input);
      assert immutableWithNulls.equals(copy);
    }
```

In this case, immutableWithNulls has both null and non-null values, so kryo catches that there are null values in a supposedly nonnull container during serialization
```
  }
  @Test
  public void testImmutableSerialization() throws Exception {
    final Kryo kryo = new Kryo();
    kryo.setRegistrationRequired(false);
    final Output output = new Output(64);
    final List<Integer> immutableWithNulls = Stream.of(null,1, null).toList();
    kryo.writeClassAndObject(output, immutableWithNulls);
    try(Input input = new Input(output.getBuffer())) {
      final Object copy = kryo.readClassAndObject(input);
      assert immutableWithNulls.equals(copy);
    }
```
**Environment:**
 - OS: Ubuntu running in WSL
 - JDK 21 (but Stream::toList was added in 16. I have no clue if that is when it started returning a ListN)
 - Kryo 5.5

**Additional context**

konlue:
@theigl I was looking at this with the intent of submitting a PR, and I think one of the two options you outlined hits a concrete blocker:

[...]

PR #1302, opened 2026-08-25
Wait for Kryo 6: ImmutableCollections ListN nulls

arimu1:
## Summary

This PR is **not** a Kryo 5 production change.

`Stream.toList()` can produce `ImmutableCollections$ListN` instances that contain nulls, but Kryo 5 cannot adopt that reconstruction path:

- Kryo 5 max source level is JDK 11 (`Stream.toList` is JDK 16+).
- Flipping `setElementsCanBeNull` is not backwards compatible.

Per review on this PR, the production fix waits for Kryo 6. The Kryo 5 serializer/tests were reverted to `master` (`setElementsCanBeNull(false)`, `List.of` / `List.copyOf`). JMH numbers for `List.of` vs `Stream.toList` are in the thread: https://github.com/EsotericSoftware/kryo/pull/1302#issuecomment-5419868460

Related: #1239

## Test plan

- [x] Reverted `JdkImmutableListSerializer` and `ImmutableCollectionsSerializersTest` to upstream `master`
- [x] Removed `ImmutableListFactoryBenchmark` from this tree (JDK 11 `mvn install` must not compile `Stream.toList`)


theigl:
Thanks for the PR @arimu1. Unfortunately, this doesn't work for 2 reasons:

- The maximum source level for Kryo 5 is JDK 11 where the toList method on stream does not exist
- The change is not backwards compatible because of the change to the `setElementsCanBeNull` flag

We will have to wait for Kryo 6 to merge this.

Also I'd be interested to know how this affects performance. Can you do a benchmark (preferable JMH) and compare the performance of `List.of` to `Stream.toList` across some different list sizes?

arimu1:
Thanks @theigl — agreed on both blockers.

- We will wait for Kryo 6. Kryo 5's max source level is JDK 11, so we will not force `Stream.toList()` onto this line (`toList` is JDK 16+; PR CI already fails the JDK 11 compile with `cannot find symbol: method toList()`).
- Changing `setElementsCanBeNull` is not a backwards-compatible path, so that is not a Kryo 5 fix either.

I added `ImmutableListFactoryBenchmark` in the existing JMH module (`List.of` vs `Stream.toList` at sizes 0, 1, 2, 10, 100, 1000). Local run on Temurin 21.0.12, macOS aarch64, `-f 2 -wi 3 -i 3 -t 1 -w 1s -r 1s`:

| size | `List.of` (ops/s) | `Stream.toList` (ops/s) | `List.of` / `toList` |
| ---: | ---: | ---: | ---: |
| 0 | 2.35e9 | 9.11e7 | ~26× |
| 1 | 6.24e8 | 7.77e7 | ~8.0× |
| 2 | 6.58e8 | 6.99e7 | ~9.4× |
| 10 | 1.16e8 | 4.03e7 | ~2.9× |
| 100 | 1.31e7 | 4.58e6 | ~2.9× |
| 1000 | 1.27e6 | 4.97e5 | ~2.6× |

`List.of` is faster at every size, especially 0/1/2 (empty singleton / `List12`). From size 10 the gap is about 2.5–3×. Reconstruction is cheap next to element ser/deser, but `Stream.toList` is clearly more expensive. Happy to re-run with longer forks if useful.

No production rewrite on this Kryo 5 PR.

arimu1:
Closing this Kryo 5 PR: there is no production path here.

Waiting for Kryo 6 (JDK 11 cannot compile `Stream.toList`; `setElementsCanBeNull` is not backwards compatible). JMH `List.of` vs `Stream.toList` numbers are already in the thread: https://github.com/EsotericSoftware/kryo/pull/1302#issuecomment-5419868460

theigl:
Thank you for the benchmark @arimu1! 

It is unfortunate that the stream API is so much slower. But I don't see another way to handle this. The only option would be to use reflection to check if the list allows nulls (`ListN.allowNulls`). But that would make the serializer unsafe and dependent on JDK internals. Another possible option would be to use separate serializers for empty list, `List12` and `ListN`. Empty list and `List12` (no nulls allowed) could stay on the more performant path.

I have to revisit this once main is on JDK17+.

